Issue #17, opened 2026-02-10
The first variable of a synthetic dataset should allow generation by methods other than sampling.


KoomenErina:
As @LotteVanUtrecht mentioned in #2, the users should be able to provide an alternative to sampling the first column in a synthesis. 
One possible usecase for this is to copy the first column from the original data. There might be potential to use this for multi-table synthesis in the far future.

KoomenErina:
This change does not change the default behaviour. Note that https://github.com/Synthpop-data/synthpop-py/issues/13 might be related

Issue #13, opened 2026-02-10
Synthesiser should be a sklearn transfrom


KoomenErina:
There are usecases where there are many steps before and after synthesising data. I have seen these usecases during the first hackaton and when trying to synthesise faces. When the` Synthesiser` class is a sklearn transform, it can be used in pipelines and other meta estimators provided by scikit-learn. 

The implication is that `Synthesiser` needs to have ` get_feature_names_out` and needs to inherit from `BaseEstimator` and `TransformerMixin`

In general, transforms cannot alter the number of rows. The default option should be to synthesise the same number of rows as there are original observations. A argument in the `__init__` should provide the option to alter the number of synthetic rows.

KoomenErina:
This issue will be taken up when there are more users that will benefit from this

PR #19, opened 2026-02-11
proposal for #17 and #13

KoomenErina:


src/synthpop/Synthesiser.py
```diff
@@ -27,20 +28,39 @@ def fit(self, X: pd.DataFrame, y=None) -> Self:
...
 
-        This method loops through the columns of ``X``, following ``column_order``, and calls the :py:meth:`transform` function of the synthesis method classes from ``default_syn_method`` and ``special_syn_method``.
+        This method loops through the columns of the data used for fitting, following ``column_order``, and calls the :py:meth:`transform` function of the synthesis method classes from ``default_syn_method`` and ``special_syn_method``.
         
-        For the first column, we simple generate a random sample with :py:meth:`pandas.DataFrame.sample`. For the next ones, we use their respective fitted functions. Each column is predicted
+        If ``X_syn`` is ``None``, then for the first column, we simple take a random sample with :py:meth:`pandas.DataFrame.sample`. For the next ones, we use their respective fitted models. Each column is predicted
         using the previously generated columns as features.
+        If ``X_syn`` is not ``None``, it is used as features to predict the first column. It will be included as features for all thereafter. 
+        Note that the columns in ``X_syn`` should be present in the data used for fitting.
```
LotteVanUtrecht, line 52: So X_syn is just a way to copy the first couple features from the original to the synthetic data set right? Can we also allow this as a ``special_syn_method`` instead of explicitly adding it?

KoomenErina: `x_syn` is a parameter of the transform method. Copying the original data is one of the usecases. Another usecase could be to use this in multi-table synthesis, or filtering based on the first column. See examples.md for examples of how this looks like for the user.

The deeper idea is that we can think of synthesising data as a transfromation in the sense of scikit-learn, meaning it can be fit on original data, and after fitting transforms input data to output data. Using `x_syn` as a parameter for the first columns makes the synthesis process behave more like a scikit-learn transformation. This makes it much easier for users to use the synthesis process in exsisting pipelines and processes. 

I understand that you suggest that we make 2 more synthesis methods, specially for the first columns: sampling the first column and copying the first column. The user would then use the `__init__` of this class to supply the synthesis method (with a default of sampling). Do you see any particular advantage for the user in this approach?

KoomenErina:
There has been decided to go with #26 instead of this pull request for now. The reason is that the users we know are synthpop-R users. The users in a machine-learning context are not yet known to us.  #13 is kept open to gather input from those users.

PR #22, opened 2026-02-12
6-edge case spmse

### LINKED ISSUE #6: The expected behaviour of the S_pMSE is not clear in every situation
Suppose X_org and Y_org are two categorical columns of the original dataset, and X_syn and Y_syn are  two categorical columns of the synthetic dataset. Suppose that X_org has values in {v_1,v_2,v_3} and X_syn has values in  {v_1,v_2} (missing a possible value). 
The spmse is calculated by summing over the possible values of X and Y. It is not clear that sum is over the possible values of X_syn and Y_syn or X_obs and Y_obs. Perhaps sum the union of values over?

bbonnet14:
… categories

### DIFF
--- a/docs/source/functional descriptions/S_pMSE.md
+++ b/docs/source/functional descriptions/S_pMSE.md
@@ -26,7 +26,7 @@ The computation of S_pMSE for a given pair of variables $(X, Y)$ (from the origi
 
 For each variable pair $(X, Y)$:
 - If a variable is numeric, it is discretised into at most $\text{max_bins}$ bins.
-- Missing values in either variable are replaced with the value "N.a.N."
+- Missing values in either variable are replaced with the value "N.a.N.". If that value is already used, an error is raised.
 
 After preprocessing, both variables are treated as categorical variables with a finite number of levels.
 
--- a/docs/source/functional descriptions/S_pMSE.md
+++ b/docs/source/functional descriptions/S_pMSE.md
@@ -8,7 +8,7 @@ The Standardised Propensity Mean Squared Error (S_pMSE) is a statistical utility
 
 The computation of S_pMSE requires the following inputs:
 - An original dataset X with $n_o$ rows and $p$ variables
-- A synthetic dataset with $n_s$ rows and the same $p$ variables. The order of the columns is not relevant, but the column names must be controlled.
+- A synthetic dataset with $n_s$ rows and the same $p$ variables. The order of the columns is not relevant, but there needs to be a check that the column names match between both datasets.
 - A maximum number of groups $\text{max_bins} \in \mathbb{N}$ used to discretise numeric variables.
 
 Let $X$ and $Y$ be any columns of the original dataset. The output of the S_pMSE function is a dataset containing all pairs of variables $(X, Y)$ with their corresponding S_pMSE value. Because $\text{S_pMSE}(X, Y)$ and $\text{S_pMSE}(Y, X)$ are equal, $\text{S_pMSE}(Y, X)$ is neither calculated or included in the output dataset.
@@ -54,7 +54,7 @@ Only category pairs with strictly positive expected frequency are retained for s
 
 ### 3.5 Calculation of the S_pMSE
 
-Let $k$ be the number of unique category pair $(x, y)$, for which $f_{orig}(x, y)$ or $f_{syn}(x, y)$ is not null.
+Let $k$ be the number of unique category pair $(x, y)$, for which $f_{orig}(x, y)$ or $f_{syn}(x, y)$ is not 0.
 
 The S_pMSE for the variable pair $(X, Y)$ is computed, for $k>1$:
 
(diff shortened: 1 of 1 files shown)


docs/source/functional descriptions/S_pMSE.md
KoomenErina: Is it really necessary that the columns are in the same order? Would columns having the same name suffice?

KoomenErina: perhaps using the word "bins" instead of "groups" would be better? A group means something else in mathematics, especially in combination with natural numbers.

KoomenErina: Does the value of na_label have an effect on the output? If it has no effect on the output, it doesn't need to be an input.

KoomenErina: perhaps use "N.a.N." as fixed value, and raise an error if that categorical level already exsists and the variable still had missing values?

KoomenErina, line 63: what is the value of $k$?
Is there a relation with what is written under paragraph 5.3?

docs/source/functional descriptions/S_pMSE.md
```diff
@@ -77,10 +77,14 @@ The statistic is invariant under swapping the original and synthetic datasets, s
...
 Missing values are treated as an explicit category. While this preserves information about missingness, it may inflate the influence of missing data patterns on the S_pMSE.
 
-### 5.2 Sparse joint distributions
+### 5.2 Constant variables
 
 If both variables are constant, the statistic is undefined due to division by zero. The function sends a warning, ignores the variable pair and continues with the next ones.
 
+### 5.3 Inequal number of categories
+
+If some combinations of variables exist in the original dataset, but not in the synthetic dataset, the synthetic frequency is 0. And vice-versa.
```
KoomenErina: I think you mean "combination of categories", not "combination of variables".

KoomenErina: You mention combinations of variables. Perhaps it's good to mention something about one variable missing a category value.

docs/source/functional descriptions/S_pMSE.md
KoomenErina, line 75: what happenbs if $X_{obs}$ is never missing and $X_{syn}$ is always missing?

KoomenErina (changes requested):

docs/source/functional descriptions/S_pMSE.md
```diff
@@ -8,9 +8,8 @@ The Standardised Propensity Mean Squared Error (S_pMSE) is a statistical utility
 
 The computation of S_pMSE requires the following inputs:
 - An original dataset X with $n_o$ rows and $p$ variables
-- A synthetic dataset with $n_s$ rows and the same $p$ variables, in the same column order.
+- A synthetic dataset with $n_s$ rows and the same $p$ variables. The order of the columns is not relevant, but the column names must be controlled.
```
KoomenErina: I think you mean "must be the same" instead of must be controlled

docs/source/functional descriptions/S_pMSE.md
```diff
@@ -26,8 +25,8 @@ The computation of S_pMSE for a given pair of variables $(X, Y)$ (from the origi
 ### 3.1 Preprocessing 
 
 For each variable pair $(X, Y)$:
-- If a variable is numeric, it is discretised into at most $\text{max_groups}$ bins.
-- Missing values in either variable are replaced with the explicit category defined by $\text{na_label}$
+- If a variable is numeric, it is discretised into at most $\text{max_bins}$ bins.
+- Missing values in either variable are replaced with the value "N.a.N."
```
KoomenErina: If "N.a.N." is already a category, an exception should be raised.

docs/source/functional descriptions/S_pMSE.md
```diff
@@ -55,7 +54,9 @@ Only category pairs with strictly positive expected frequency are retained for s
 
 ### 3.5 Calculation of the S_pMSE
 
-The S_pMSE for the variable pair $(X, Y)$ is computed as:
+Let $k$ be the number of unique category pair $(x, y)$, for which $f_{orig}(x, y)$ or $f_{syn}(x, y)$ is not null.
```
KoomenErina: "unique category pairs". 

Null is usually interpreted as empty/missing/undefined. Perhaps it's better to use "zero" or "0"?

For my understanding, if $X$ and $Y$ both have 2 possible values ("A" and "B"), but they only occur in the following 2 combinations:  (X="A" and Y="B") or (X="B" and Y="A"), is it true that $k$ is equal to 2?

KoomenErina (changes requested):

docs/source/functional descriptions/S_pMSE.md
bbonnet14: Is that needed? I'd rather just have more values in the missing category.

docs/source/functional descriptions/S_pMSE.md
bbonnet14: Yes that is true.

docs/source/functional descriptions/S_pMSE.md
KoomenErina: My reasoning is that it is a weird situation to have a "N.a.N." category and still missing values. I think that it indicates an error earlier on.

KoomenErina (approved):

PR #26, opened 2026-02-18
Alternative proposal #17 #13

KoomenErina:
This pull request is an alternative to #19. 
In this pull request, the process of synthesising data is not seen as a transformation in the sense of scikit-learn. 
Instead, both sampling and copying a column are seen as methods of synthesising data. Generating the first column is seen as an edgecase of the synthesis process. 

Both code and examples have been updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 /docs/build/
 /dist/
 .vscode
+
--- a/docs/source/functional descriptions/CART.md
+++ b/docs/source/functional descriptions/CART.md
@@ -28,6 +28,9 @@ The algorithm consists of two phases:
 
 How the synthesiser is fit depends on whether the target column contains a numeric or categorical variable. Below we will explain the different approaches.
 
+#### 3.1.0 The first column
+The first column is a special case for CART. If there are no predictors, a sample with replacement should be taken from the distribution of the observed data. 
+
 #### 3.1.1 Fitting with a categorical target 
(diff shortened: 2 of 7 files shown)


KoomenErina:
advantages for users of this approach (in relation to #19):
- more familiar for synthpop-r users.
- The treatment for every output column is described in the same place.
- Makes sense if you are not familiar with how synthpop roughly works.

Disadvantages for users of this approach (in relation to #19):
- The synthesis process cannot fluently be incorporated in machinelearning flows.
- Combining synthpop with other data processes is more difficult.
- If you focus on the flow of the data, it is not immediatly clear what data is purely a predictor.
- Copying columns as a method of synthesising data might mislead the user in terms of privacy.

Neither this pull request nor #19 solves the usecase for the predictor matrix in synthpop-R. 

This pull requests does not highlight the predictor/target structure inherit to machine learning. The columns that are not being synthesised are being "copied". Pull request #19 highlights the predictor/target structure. The columns that are not being synthesised are used only as predictor.

KoomenErina:
There are also a few design/maintainability considerations:
- The CopyMethod is not changing anything. Having code with as purpose "doing nothing" makes the design more complex.
- Making the method for the first column modular in the same way as for other columns complicates the hierarchy of classes.  This is on it's own not a major problem, but makes the design conceptually more complex. The reason for this extra complexity is that the requirements for a synthesis method for the first column are different than the requirements for other synthesis methods.
- If sampling or copying are the only reasonable options, then the added value of  the more complicated design is low. However, if there are more options that a significant amount of users would use, it might way up against the more complicated design.

KoomenErina:
There has been decided to go with this pull request instead of #19 for now. The reason is that the users we know are synthpop-R users. The users in a machine-learning context are not yet known to us.  #13 is kept open to gather input from those users.

LotteVanUtrecht:
Approved!

PR #27, opened 2026-02-18
25-Create new mean encoder

### LINKED ISSUE #25: Add mean encoder
Requirements of the MeanEncoder class are specified in the technical description.

bbonnet14:
This pull request closes issue #25. The purpose is to test the implementation of the technical description. Edge cases should be included and properly tested. 
The mean encoder is meant to be used in the context of a synthesis, but should also function independently.


KoomenErina:
The docs don't build. 
The following commands didn't work:
```bash
make html
poetry run sphinx-build -M  html "source" "build"
```

KoomenErina:
the unit tests are in the file "test_encoders.py". This implies that the unit tests for the PCA encoder will also be in this file. I think it would be better to name this file "test_mean_encoder.py"?

tests/unit/test_encoders.py
```diff
@@ -0,0 +1,126 @@
...
+
+    encoder = MeanEncoder()
+    encoder.fit(X, y)
+
+    X_transformed = encoder.transform(X)
+    assert X_transformed.equals(pd.DataFrame({'color':[np.nan, 1, np.nan]})), "If y is always empty for a specific X category, encoding must be empty."
+
+def test_some_missing_target_values_are_ignored():
+    X = pd.Series(["red", "blue", "red", "blue", "red"], name='color')
+    y = pd.Series([1, 0, 2, np.nan, np.nan], name='score')
```
KoomenErina: In these test you only use np.Nan. There should be test that include None as well

tests/unit/test_encoders.py
```diff
@@ -0,0 +1,126 @@
...
+    assert encoder.mapping_['blue'] == 0
+
+def test_is_fitted():
+    X = pd.Series(["red", "blue", "red", "blue", "red"], name='color')
+    y = pd.Series([1, 0, 2, 0, 3], name='score')
+
+    encoder = MeanEncoder()
+    encoder.fit(X, y)
+
+    assert any(encoder.mapping_)
```
KoomenErina: this test is not needed when you have `test_calculate_means`.  
The test `test_calculate_means` implies that `encoder.mapping_` exsists and is not empty for exactly the same inputs. You can remove `test_is_fitted` without losing any coverage. You could add this assertion (`assert any(encoder.mapping_)`) to test_calculate_means.

tests/unit/test_encoders.py
```diff
@@ -0,0 +1,126 @@
...
+    Z = pd.Series(["red", "blue", "green"], name='color')
+    with pytest.raises(ValueError):
+        encoder.transform(Z)    
+
+def test_get_feature_names_out():
+    X = pd.Series(["red", "blue"], name="color")
+    y = pd.Series([1, 0], name="score")
+
+    encoder = MeanEncoder()
+    names = encoder.fit(X, y).get_feature_names_out()
```
KoomenErina: This is the only part in the test where you (implicitly) verify the return value of encoder.fit().
In this test, you only verify that the return value is an object with a `get_feature_names_out` method, which works correctly. 

The functional descripions and typehints specify that fit should return `self`.  That is something that needs to be added somewhere in the tests.

tests/unit/test_encoders.py
```diff
@@ -0,0 +1,126 @@
...
+    X = pd.Series(["red", "blue", "red", "blue", "red"], name='color')
+    y = pd.Series([1, 0, 2, np.nan, np.nan], name='score')
+
+    encoder = MeanEncoder()
+    encoder.fit(X, y)
+
+    X_transformed = encoder.transform(X)
+    assert X_transformed.equals(pd.DataFrame({'color':[1.5, 0, 1.5, 0, 1.5]})), "If some values are missing for a specific X category, they are ignored."
+
+def test_no_error_with_constant_feature():
```
KoomenErina: The name of the test is not descriptive enough. "no error" is not a good description of desired behavior. It is better to write what is expected given this scenario.
The desired behaviour in this case is a constant encoding.

tests/unit/test_encoders.py
```diff
@@ -0,0 +1,126 @@
...
+    X = pd.Series(["red", "red", "red"], name='color')
+    y = pd.Series([1, 2, 3], name='score')
+
+    encoder = MeanEncoder()
+    encoder.fit(X, y)
+
+    X_transformed = encoder.transform(X)
+    assert X_transformed.equals(pd.DataFrame({'color':[2.0, 2.0, 2.0]})), "Constant feature does not trigger error"
+
+def test_no_error_with_constant_target():
```
KoomenErina: see test_no_error_with_constant_feature():

KoomenErina:
There are only tests where the target is nummeric. One test where the target has values like "1/3" might highlight subtle floatingpoint issues.

pyproject.toml
```diff
@@ -26,9 +26,13 @@ docs = [
     "sphinxcontrib-mermaid (>=2.0.0,<3.0.0)",
 ]
 dev = [
+    "pytest (>=9.0.2,<10.0.0)",
     "pytest-cov (>=7.0.0,<8.0.0)",
     "pylint (>=4.0.4,<5.0.0)",
 ]
+[tool.poetry.group.docs]
+optional = true
+
```
KoomenErina, line 35: Why has this been changed?

KoomenErina:
In these test, you test the fit method together with the transform method. A bug in either of them can make a test fail. Especially for more complicated components, this becomes a problem

tests/unit/test_encoders.py
bbonnet14: I removed test_is_fitted and added this assertion in test_calculate_means

tests/unit/test_encoders.py
bbonnet14: Added None value.

tests/unit/test_encoders.py
bbonnet14: renamed test_constant_encoding_by_constant_feature

tests/unit/test_encoders.py
bbonnet14: Renamed test_constant_encoding_with_constant_target

tests/unit/test_encoders.py
bbonnet14: Added test_fit_returns_self

pyproject.toml
bbonnet14: That came automatically with the installation of pytest.

bbonnet14:
I can't directly reply to your comments that are not linked to the code, so here are my responses:
- I changed the test file's name
- I added a test test_calculate_means_with_fractional_value, can you take a look?
- The transform does not work without the fit, how can I test it otherwise?
- The documentation runs for me with these steps: 
  1. Open a terminal in the root of the repository:
  2. Navigate to the **docs** directory: ```cd docs``` 
  3. Delete current documentation files: ``` poetry run sphinx-build -M clean source build```
  4. Build the new documentation: ``` poetry run sphinx-build source build```

tests/unit/test_mean_encoder.py
```diff
@@ -0,0 +1,138 @@
...
+
+def test_get_feature_names_out():
+    X = pd.Series(["red", "blue"], name="color")
+    y = pd.Series([1, 0], name="score")
+
+    encoder = MeanEncoder()
+    names = encoder.fit(X, y).get_feature_names_out()
+
+    assert isinstance(names, np.ndarray), "Output must be an numpy array"
+    assert names.tolist() == ["color"], "Output has incorrect name"
```
KoomenErina: There is no test to assert that when the input is None, the output is None as well.

KoomenErina:
> I can't directly reply to your comments that are not linked to the code, so here are my responses:
> 
>     * I changed the test file's name
> 
>     * I added a test test_calculate_means_with_fractional_value, can you take a look?
> 
>     * The transform does not work without the fit, how can I test it otherwise?
> 
>     * The documentation runs for me with these steps:
>       
>       1. Open a terminal in the root of the repository:
>       2. Navigate to the **docs** directory: `cd docs`
>       3. Delete current documentation files: ` poetry run sphinx-build -M clean source build`
>       4. Build the new documentation: ` poetry run sphinx-build source build`

The only reason why `fit()` needs to be called before `transform` is because transform needs the `mapping_` property. 
You could use one set of tests to verify that `mapping_` is set correctly when fitting, and another set of tests to verify that `transform` has the correct output given that `mapping_` has been set correctly. You can make that second set of tests by using `encoder.mapping_ = {"a":1,"b":2}` in the unit tests. This would simplify your tests as well. 

The advantage of this approach is that the expected state of the encoder object before calling `transform` becomes clear from the tests. The change of state when `fit()` is called becomes clear too. For now, the impact of not doing this is small because the requirements are relatively simple. But if other ways of obtaining a fitted estimator become available (for example, read from file when continuing a synthesis) or when `fit` becomes more complex, the difference starts to matter.

src/synthpop/data_processing/Encoders.py
```diff
@@ -40,11 +42,58 @@ def __init__(self):
...
+        Examples
+        X = pd.Series(["red", "blue", "red", "blue", "red"], name='color')
+        y = pd.Series([1, 0, 2, 0, 3], name='score')
+
+        encoder = MeanEncoder()
+        encoder.fit(X, y)
+        """
+        # Required for get_feature_names_out
+        self.feature_names_in_ = np.array([X.name], dtype=object)
+        self.n_features_in_ = 1
```
KoomenErina, line 60: not needed to do this ourselves, see https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.validate_data.html

KoomenErina (changes requested):

src/synthpop/data_processing/Encoders.py
bbonnet14: I tried to apply validate_data but it seems that it does not work if X is categorical

bbonnet14:
@KoomenErina I merged your review branch, so if you don't have further feedback, feel free to merge this one into develop :)

KoomenErina (approved):

PR #36, opened 2026-02-19
34 add licence and funding

KoomenErina:


LotteVanUtrecht (commented):
Do we want to add funding here?

KoomenErina:
> Do we want to add funding here?

not yet I think

PR #43, opened 2026-02-24
39 implement pca encoding

KoomenErina:
implemented PCA encoding.

it had to be reimplemented because of #45. 

The large number of commits is due to merging without squashing. 
The focus of the review should be the following files:
 - encoders.py
 - test_pca_encoder.py
 - test_int_pca_encoding.py


src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
+
+
+class PCAEncoder(TransformerMixin, BaseEstimator):
     """
     Transforms categorical data to one or more numeric columns.
+    The user can adjust the amount of principle components by passing an instance of sklearn.decomposition.PCA to ``_pca_transform``
+
+    :param _pca_transform: The pca transform used. See `sklearn.decomposition.PCA <https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html/>`_ for the possible parameters.
+
+    Examples
```
anneliesdej, line 25: I would change one of the examples so it shows how to adjust the amount of principle components as well

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
 
-class PCAEncoder(TransformerMixin, BaseEstimator): 
+
+
+class PCAEncoder(TransformerMixin, BaseEstimator):
     """
     Transforms categorical data to one or more numeric columns.
+    The user can adjust the amount of principle components by passing an instance of sklearn.decomposition.PCA to ``_pca_transform``
+
+    :param _pca_transform: The pca transform used. See `sklearn.decomposition.PCA <https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html/>`_ for the possible parameters.
```
anneliesdej: Mention the default here?

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
-    def __init__(self, PCA_threshold:int = 30, explained_variance:float = 0.95):
-        pass
+    def __sklearn_tags__(self):
+        tags = super().__sklearn_tags__()
+        tags.target_tags.required=False
+        tags.target_tags.one_d_labels = True
+        tags.target_tags.single_output= True
+
+        tags.input_tags.categorical = True
+        #tags.input_tags.two_d_array = False
```
anneliesdej: for other developers it might be handy to explain why it is commented out instead of just removed

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
-        :param X: The categorical feature that is to be encoded.
-        :param y: The target used to encode the feature.
+        Calculate the encoding.
 
-        :return: fitted encoder.
+        :param X: 1D array of categorical data. This is contains the data to be encoded.
+        :param y: 1D array of categorical data. The encoding is based on this data.
         """
+
+        self.n_features_in_ = 1
```
anneliesdej, line 106: for code readability it might be nice to first have a comment line with something like # data input validation checks, then all the code that belongs to that and then later another comment line when you start the actual implementation

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
+        input_feature_pca0  input_feature_pca1  input_feature_pca2
+        0            1.443766           -0.163255        2.117578e-17
+        1            1.443766           -0.163255        2.117578e-17
+        2           -0.932313           -0.758443        2.117578e-17
+        3           -0.932313           -0.758443        2.117578e-17
+        4           -0.511452            0.921698        2.117578e-17
+    
+    """
 
+    def __init__(self, _pca_transform:PCA = PCA()):
```
anneliesdej: From scikit-learn developer:
"To summarize, an __init__ should look like: There should be no logic, not even input validation, and the parameters should not be changed; which also means ideally they should not be mutable objects such as lists or dictionaries. If they’re mutable, they should be copied before being modified. The corresponding logic should be put where the parameters are used, typically in fit."
In interpret this as that PCA() is mutable because it is a fitted object so it should look like __init__(self, pca_transform=None) and then in fit:
if self.pca_transform is none: self._pca_transform_ = PCA() else: self._pca_transform_ = clone(self.pca_transform)

Additionally, pconstructer parameters should not start with _, learned parameters can

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
+        self._pca_transform = _pca_transform
 
-    :param PCA_threshold: maximum number of columns used to encode the feature. explained_variance_threshold has precedence above PCA_threshold.
-    :param explained_variance: parameter indicating how much of the total variance should be explained by the principle components. A value of 1 returns all principle components.
-    """
-    def __init__(self, PCA_threshold:int = 30, explained_variance:float = 0.95):
-        pass
+    def __sklearn_tags__(self):
+        tags = super().__sklearn_tags__()
+        tags.target_tags.required=False
```
anneliesdej: fit() requires y so this tag should be True right?

src/synthpop/data_processing/encoders.py
```diff
@@ -37,10 +135,30 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
+        # In that case, X cannot be sorted.
+        # many routines of numpy for finding differences between lists depend on the items being sortable.
+        unique_values_in_x = np.unique([str(v) for v in X if v is not None])
+
+        if len(np.setdiff1d(unique_values_in_x,list(mapping_including_missing.keys()),assume_unique=True)) !=0:
+            raise ValueError("new values not seen during fitting when encoding.")
+
+        return np.array(
+            [# if the categorical data is represented as integers (as floats) (as in the standard sklearn tests), floating point errors can emerge when indexing the dictionary.
+                mapping_including_missing[str(v) if v is not None else None] for v in X
```
anneliesdej: this only works if the original values are strings. If they are integers the lookup fails

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
+        # that means that the user might configure globally that all transformers return pandas dataframes.
+        if isinstance(pca_result,pd.DataFrame):
+            pca_result = pca_result.to_numpy()
+
+        self.n_features_out_ = pca_result.shape[1] #needed for get_feature_names_out
+
+        value_mapping = {contingency_table.index[i]: pca_result[i] for i in range(pca_result.shape[0])}
+
+        #The alternative to using pandas here is either use scipy or DIY.
+        missing_contingency_table = pd.crosstab(X_val,[v is None or v is pd.NA or v is np.nan for v in y_val])
```
anneliesdej: `is` is not correct for NumPy scalars. and pd.isna(y_val) is recommended instead of pd.NA. pd.isna(y_val) works for array-like objects

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
+            raise ValueError("X should by 1D")
+        if y_val.ndim != 1:
+            raise ValueError("Y should by 1D")
+   
+        #The alternative to using pandas here is either use scipy or DIY.
+        contingency_table = pd.crosstab(X_val,y_val,) #the result of pd.crosstab is a pandas dataframe
+
+        self._pca_transform_ = clone(self._pca_transform)# for compatibility with sklearn we need to do this.
+
+        pca_result = self._pca_transform_ .fit_transform(
```
anneliesdej: sklearn.PCA does not scale columns by variance, wihch is mentioned in the functional description as step 3.3

src/synthpop/data_processing/encoders.py
```diff
@@ -37,10 +135,30 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
 
         :param X: the feature to be encoded.
         """
-        return pd.DataFrame()
-    
-    def get_feature_names_out(self):
-        pass
+
+        mapping_including_missing = self.mapping_| {None:[None]*self.n_features_out_}
```
anneliesdej: does this work correctly for np.nan as well? Since that is not None

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,131 @@
...
+
+        if isinstance(X,pd.Series):#validate data does not seem to get the name of the feature when it is a pd.Series instead of a pd.Dataframe.
+            self.feature_names_in_ = [X.name]
+        if X_val.ndim != 1:
+            raise ValueError("X should by 1D")
+        if y_val.ndim != 1:
+            raise ValueError("Y should by 1D")
+   
+        #The alternative to using pandas here is either use scipy or DIY.
+        contingency_table = pd.crosstab(X_val,y_val,) #the result of pd.crosstab is a pandas dataframe
```
anneliesdej: Right now it seems like when the target is constant the PCA reduces to count encoding as sklearn.PCA centres which substracts the mean so results are not counts but counts minus mean (functional description). For a constant feature it produces a zero vector, where the functiona description mentiones that the encoding should be the total number of rows

tests/unit/test_pca_encoder.py
```diff
@@ -0,0 +1,385 @@
...
+    assert len(result.mapping_["c"]) == 2
+    assert result.mapping_["c"].count(None) == 2
+
+    validate_set_inout_count(result,expected_n_feat=2)
+
+@parametrize_with_checks([PCAEncoder()],legacy=False,expected_failed_checks= lambda x: {
+    "check_dont_overwrite_parameters":"tests with multiple features",
+    "check_n_features_in_after_fitting":"tests with multiple features"
+})
+def test_pca_encoder_is_sklearn_compatible(estimator,check):
```
anneliesdej: two tests get skipped in this test suit (check_dont_overwrite_parameters and check_n_features_in_after_fitting)

tests/integration/test_int_pca_encoding.py
```diff
@@ -0,0 +1,55 @@
...
+
+    encoder = PCAEncoder()
+
+    encoder.fit(X=X,y=y)
+
+    assert len(encoder.mapping_) == 3
+    assert len(encoder.mapping_["b"]) == 1
+    assert encoder.mapping_["a"][0] == pytest.approx(encoder.mapping_["b"][0])
+
+    # The behaviour is different as described in the functional descriptions.
```
anneliesdej, line 59: See my comment on this a well. Either the code or functional description should be changed

tests/unit/test_pca_encoder.py
```diff
@@ -0,0 +1,385 @@
...
+
+def test_pca_transform_exception_on_new_value():
+    """
+    An informative exception should be raised when the user attempts to encode a value that was not in the data during fitting.
+    """
+    encoder = PCAEncoder(_pca_transform=None)
+    encoder.mapping_={"a":[1.1,2.2]}
+    encoder.n_features_out_=2
+
+    with pytest.raises(ValueError,match="values not seen during fitting"):
```
anneliesdej: This works, but the exact error message is "new values not seen during fitting when encoding."

tests/integration/test_int_pca_encoding.py
```diff
@@ -0,0 +1,55 @@
+import pytest
```
anneliesdej, line 1: There are currently no tests on numeric correctness, transform correctness or variance scaling

tests/unit/test_pca_encoder.py
```diff
@@ -0,0 +1,385 @@
+import pandas as pd
```
anneliesdej, line 1: There is no test for mathematical properties such as the rank (func desc), no variance scaling and integer categories as keys. To do also mentions the test that first should be fitted before transform can be called (sklearn guide NotFittedError). Maybe even a real PCA numeric verification test with like 2x2 contingency

anneliesdej (changes requested):

src/synthpop/data_processing/encoders.py
KoomenErina: done

src/synthpop/data_processing/encoders.py
KoomenErina: comment removed

src/synthpop/data_processing/encoders.py
KoomenErina: done

src/synthpop/data_processing/encoders.py
KoomenErina: done

tests/unit/test_pca_encoder.py
KoomenErina: it does not need to be the exact error message. I wanted to test if the most important part of the error was in there.

tests/integration/test_int_pca_encoding.py
KoomenErina: I am not sure that a test for numeric correctness is needed here. The actual heavy math happens in `sklearn.decomposition.pca`. We only need to test that PCA is called correctly, and assume that the implementation of PCA by sklearn is correct. The unit tests prove that the output of PCA is processed correctly

src/synthpop/data_processing/encoders.py
KoomenErina: true

src/synthpop/data_processing/encoders.py
KoomenErina: There are tests covering this. Do you agree?

src/synthpop/data_processing/encoders.py
KoomenErina: solved

tests/unit/test_pca_encoder.py
KoomenErina: true. Those test don't seem to work with 1D data, and seem to fail for reasons not related to the code.

tests/integration/test_int_pca_encoding.py
KoomenErina: solved

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,164 @@
...
-        :return: fitted encoder.
+        :param X: 1D array of categorical data. This is contains the data to be encoded.
+        :param y: 1D array of categorical data. The encoding is based on this data.
         """
+
+        self.n_features_in_ = 1
+
+        # Input validation
+
+        if X.shape[0] == 0 and y.shape[0]==0:#validate_data does not work well when there is no data
```
anneliesdej: it would be better to let validate_data raise the error by setting ensure_min_samples

docs/source/functional descriptions/PCA-encoding.md
anneliesdej: variance needs to be changed to standard deviation

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,164 @@
...
+            return self
+
+        X_val,y_val = validate_data(self,X=X,y=y, validate_separately = (
+            dict(ensure_2d=False,dtype=["str","object"],ensure_all_finite="allow-nan")
+            ,dict(ensure_2d=False,dtype=["str","object"],ensure_all_finite="allow-nan")
+            ))
+
+        if isinstance(X,pd.Series):#validate data does not seem to get the name of the feature when it is a pd.Series instead of a pd.Dataframe.
+            self.feature_names_in_ = [X.name]
+        if X_val.ndim != 1:
```
anneliesdej, line 115: this is handled in validate_data

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,164 @@
...
+            raise ValueError("X should by 1D")
+        if y_val.ndim != 1:
+            raise ValueError("Y should by 1D")
+
+        # the core of this implementation
+
+
+        #The alternative to using pandas here is either use scipy or DIY.
+        contingency_table = pd.crosstab(X_val,y_val,) #the result of pd.crosstab is a pandas dataframe
```
anneliesdej: the functional description mentions that missing values are handled as normal categories, here it should then be dropna=False

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,164 @@
...
+
+        pca_input = scale(contingency_table,axis=0)
```
anneliesdej, line 146: right now if the feature has only one non-missing level, scaling will give an input matrix of only zeroes. Depending on the PCA solver this can give issues (in combination with specification of n_components since it must be <= min(n_samples, n_features). The functional description states that if a feature column is constant the encoding becomes the total number of rows / zero vector. We should add this deterministically so it does not depend on the PCA solver

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,164 @@
...
+        # that means that the user might configure globally that all transformers return pandas dataframes.
+        if isinstance(pca_result,pd.DataFrame):
+            pca_result = pca_result.to_numpy()
+
+        self.n_features_out_ = pca_result.shape[1] #needed for get_feature_names_out
+
+        value_mapping = {contingency_table.index[i]: pca_result[i] for i in range(contingency_table.shape[0])}
+
+        #The alternative to using pandas here is either use scipy or DIY.
+        missing_contingency_table = pd.crosstab(X_val,[v is None or pd.isna(v) or v == np.nan for v in y_val])
```
anneliesdej: v == np.nan is always False because of floating point rules. is None is redundant when pd.isna is used. pd.isna already works vectorised on arrays as well so just pd.crosstab(X_val, pd.isna(y_val)) will work

src/synthpop/data_processing/encoders.py
```diff
@@ -37,12 +168,33 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
+        # In that case, X cannot be sorted.
+        # many routines of numpy for finding differences between lists depend on the items being sortable.
+        unique_values_in_x = np.unique([str(v) for v in X if v is not None])
+
+        if len(np.setdiff1d(unique_values_in_x,list(mapping_including_missing.keys()),assume_unique=True)) !=0:
+            raise ValueError("new values not seen during fitting when encoding.")
+
+        return np.array(
+            [# if the categorical data is represented as integers (as floats) (as in the standard sklearn tests), floating point errors can emerge when indexing the dictionary.
+                mapping_including_missing[str(v) if v is not None else None] for v in X
```
anneliesdej: Additionally, this won't recognise np.nan as np.nan is None gives False. So safer lookup is key = None if pd.isna(v) else v
mapping_including_missing[key]

src/synthpop/data_processing/encoders.py
anneliesdej: In the test I can only find:
def test_pca_transform_given_fitted_estimator_when_transforming_missing_values():
    X = np.array(["a", None])
Which does not test for np.nan in transform. Fit does test for np.nan

tests/unit/test_pca_encoder.py
anneliesdej: can these be removed or ignored in a way, so it cannot raise confusion?

anneliesdej (changes requested):

tests/unit/test_pca_encoder.py
KoomenErina: as far as I know this is the way to ignore those tests.

anneliesdej: I just read up on it. You are correct

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,161 @@
...
 
-class PCAEncoder(TransformerMixin, BaseEstimator): 
+
+
+class PCAEncoder(TransformerMixin, BaseEstimator):
     """
     Transforms categorical data to one or more numeric columns.
+    The user can adjust the amount of principle components by passing an instance of sklearn.decomposition.PCA to ``pca_transform``
+
+    :param pca_transform: The pca transform used. See `sklearn.decomposition.PCA <https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html/>`_ for the possible parameters.
```
anneliesdej: I meant, mention that PCA from sklearn is the default, since, technically, others can be used as well

src/synthpop/data_processing/encoders.py
anneliesdej: the code still needs a fallback here. Like: if contingency_table.shape[0] == 1:
    self.mapping_ = {contingency_table.index[0]: np.zeros(1)}
    return self

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,161 @@
...
+
+        X_val,y_val = validate_data(self,X=X,y=y, validate_separately = (
+            dict(ensure_2d=False,dtype=["str","object"],ensure_all_finite="allow-nan",ensure_min_samples=0)
+            ,dict(ensure_2d=False,dtype=["str","object"],ensure_all_finite="allow-nan",ensure_min_samples=0)
+            ))
+
+        if isinstance(X,pd.Series):#validate data does not seem to get the name of the feature when it is a pd.Series instead of a pd.Dataframe.
+            self.feature_names_in_ = [X.name]
+        if X_val.ndim != 1:
+            raise ValueError("X should by 1D")
```
anneliesdej: typos

src/synthpop/data_processing/encoders.py
```diff
@@ -37,12 +165,33 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
+        if len(np.setdiff1d(unique_values_in_x,list(self.mapping_.keys()),assume_unique=True)) !=0:
+            raise ValueError("new values not seen during fitting when encoding.")
+
+        return np.array(
+            [# if the categorical data is represented as integers (as floats) (as in the standard sklearn tests), floating point errors can emerge when indexing the dictionary.
+                mapping_including_missing[str(v) if not pd.isna(v) else None] for v in X
+            ],dtype=np.float32
+        )
+
+    def get_feature_names_out(self,input_features=None):
```
anneliesdej, line 202: Sklearn mentions that if feature_names_in_ is not set in fit then the names should be x0, x1, ..., which of course can be x0_pca, x1_pca:
"Primarily for [feature extractors](https://scikit-learn.org/stable/glossary.html#term-feature-extractors), but also used for other transformers to provide string names for each column in the output of the estimator’s [transform](https://scikit-learn.org/stable/glossary.html#term-transform) method. It outputs an array of strings and may take an array-like of strings as input, corresponding to the names of input columns from which output column names can be generated. If input_features is not passed in, then the feature_names_in_ attribute will be used. If the feature_names_in_ attribute is not defined, then the input names are named [x0, x1, ..., x(n_features_in_ - 1)]."

src/synthpop/data_processing/encoders.py
```diff
@@ -37,12 +165,33 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
+
+        check_is_fitted(self)
+        mapping_including_missing = self.mapping_| {None:[np.nan]*self.n_features_out_, pd.NA:[np.nan]*self.n_features_out_}
+
+        # if X contains Nones, then the dtype of X is object.
+        # In that case, X cannot be sorted.
+        # many routines of numpy for finding differences between lists depend on the items being sortable.
+        unique_values_in_x = np.unique([str(v) for v in X if not pd.isna(v)])
+
+        if len(np.setdiff1d(unique_values_in_x,list(self.mapping_.keys()),assume_unique=True)) !=0:
```
anneliesdej: np.in1d might compute faster as it does not sort. [python set diff](https://github.com/numpy/numpy/issues/21804) could also potentially be faster

tests/unit/test_pca_encoder.py
```diff
@@ -0,0 +1,398 @@
...
+
+def validate_set_inout_count(result_encoder, expected_n_feat):
+    assert result_encoder.n_features_in_ == 1
+    assert result_encoder.n_features_out_ == expected_n_feat
+
+    if hasattr(result_encoder, "pca_transform_"):
+        # to be compatible with sklearn.
+        assert result_encoder.pca_transform is not result_encoder.pca_transform_
+
+# test fitting pca encoder ------------------------------------------------------------------------
```
anneliesdej, line 189: a test can be considered where it checks if fit handles incorrect types

tests/unit/test_pca_encoder.py
```diff
@@ -0,0 +1,398 @@
...
+    with pytest.raises(ValueError):
+        encoder.fit(np.array(["a", None, "b", "b"]),
+                    np.array([["a", None, "b", "b"]]))
+
+
+# testing transform--------------------------------------------------------------------------------
+
+
+def get_test_data_transform():
+    missing_types = [None, np.nan]# pd.NA is not included, since pd.NA is neither numeric nor object dtype
```
anneliesdej: technically pd.NA can be included in X for transform so should be tested

src/synthpop/data_processing/encoders.py
```diff
@@ -37,12 +165,33 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
-    def get_feature_names_out(self):
-        pass
+
+        check_is_fitted(self)
+        mapping_including_missing = self.mapping_| {None:[np.nan]*self.n_features_out_, pd.NA:[np.nan]*self.n_features_out_}
+
+        # if X contains Nones, then the dtype of X is object.
+        # In that case, X cannot be sorted.
+        # many routines of numpy for finding differences between lists depend on the items being sortable.
+        unique_values_in_x = np.unique([str(v) for v in X if not pd.isna(v)])
```
anneliesdej: this lookup needs to be tested on numeric vs string colision to prevent 'invisible' bugs

tests/unit/test_pca_encoder.py
```diff
@@ -0,0 +1,398 @@
...
+    y = np.array(["x", "x", "y", "z"])
+
+    expected_input_pca = np.array([  # centred table       contingency table
+        # x y z                             x  y  z              x  y  z
+        [0, 0, 0],  # a                      a [0, 0, 0]            [2, 1, 1]
+    ])
+    return [(X, y, expected_input_pca, expected_input_pca, {"a": [0, 0, 0]})]
+
+
+def get_test_data_target_constants():
```
anneliesdej, line 103: constant feature constant target combination can also be added to tests to check if confirmed output is np..nan*n_features_out_

tests/unit/test_pca_encoder.py
KoomenErina: it can be included in X, and that is tested ( ` missing_x in [None,pd.NA]`), line 289. 
Line 265 (`missing_types = [None, np.nan]`) is about missing types in the `mapping_`. And `pd.NA` as value in a mapping (so something like `mapping_ = {"a":[pd.NA]}` would be considered a bug as the mapping should result in numeric values, which `pd.NA` is not.

I will make this more clear in the tests

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,168 @@
...
+
+
+        #The alternative to using pandas here is either use scipy or DIY.
+        missing_contingency_table = pd.crosstab(X_val,[pd.isna(v) for v in y_val])     
+        x_such_that_y_is_not_always_missing = missing_contingency_table[missing_contingency_table[False]!=0].index
+        
+        contingency_table = pd.crosstab(X_val,y_val,dropna=False).loc[x_such_that_y_is_not_always_missing]#the result of pd.crosstab is a pandas dataframe
+
+        if contingency_table.shape[0] == 1:
+            print(y)
```
anneliesdej: These print statements should be deleted before merging

src/synthpop/data_processing/encoders.py
```diff
@@ -37,11 +172,44 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
+            return np.zeros(X.shape[0])
+        mapping_including_missing = self.mapping_| {None:[np.nan]*self.n_features_out_, pd.NA:[np.nan]*self.n_features_out_}
+
+        # if X contains Nones, then the dtype of X is object.
+        # In that case, X cannot be sorted.
+        # many routines of numpy for finding differences between lists depend on the items being sortable.
+        unique_values_in_x = np.unique([str(v) for v in X if not pd.isna(v)])
+        
+
+        if np.isin(unique_values_in_x,list(self.mapping_.keys()),assume_unique=True,invert=True).all():
```
anneliesdej: seems to be an if all values are unseen instead of if any values are unseen. Think input ["a", "b"] mapping {"a": 1} no error, input ["b"] same mapping, error. Mixed input with seen and new will error test test_pca_transform_exception_on_new_value()

src/synthpop/data_processing/encoders.py
```diff
@@ -37,11 +172,44 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
+
+        check_is_fitted(self)
+        if pd.isna(X).all():
+            return np.zeros(X.shape[0])
+        mapping_including_missing = self.mapping_| {None:[np.nan]*self.n_features_out_, pd.NA:[np.nan]*self.n_features_out_}
+
+        # if X contains Nones, then the dtype of X is object.
+        # In that case, X cannot be sorted.
+        # many routines of numpy for finding differences between lists depend on the items being sortable.
+        unique_values_in_x = np.unique([str(v) for v in X if not pd.isna(v)])
```
anneliesdej: str() can be removed

src/synthpop/data_processing/encoders.py
```diff
@@ -37,11 +172,44 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
         :param X: the feature to be encoded.
         """
-        return pd.DataFrame()
-    
-    def get_feature_names_out(self):
-        pass
+
+        check_is_fitted(self)
+        if pd.isna(X).all():
+            return np.zeros(X.shape[0])
```
anneliesdej: This might need to be reshaped. Sklearn expects a transform output (n_samples, n_features_out_) and this is (n_samples,). But I think I did the same with mean encoder

src/synthpop/data_processing/encoders.py
```diff
@@ -37,11 +172,44 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
+        mapping_including_missing = self.mapping_| {None:[np.nan]*self.n_features_out_, pd.NA:[np.nan]*self.n_features_out_}
```
anneliesdej: I think it is better to explicitly handle np.nan as well, to be sure

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,168 @@
...
+            ))
+
+        if isinstance(X,pd.Series):
+            self.feature_names_in_ = [X.name]
+        if X_val.ndim != 1:
+            raise ValueError("X should be 1D")
+        if y_val.ndim != 1:
+            raise ValueError("Y should be 1D")
+        
+        if X.shape[0] == 0 and y.shape[0]==0:
```
anneliesdej: errors on list input

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,168 @@
...
+
+        #The alternative to using pandas here is either use scipy or DIY.
+        missing_contingency_table = pd.crosstab(X_val,[pd.isna(v) for v in y_val])     
+        x_such_that_y_is_not_always_missing = missing_contingency_table[missing_contingency_table[False]!=0].index
+        
+        contingency_table = pd.crosstab(X_val,y_val,dropna=False).loc[x_such_that_y_is_not_always_missing]#the result of pd.crosstab is a pandas dataframe
+
+        if contingency_table.shape[0] == 1:
+            print(y)
+            print(contingency_table)
```
anneliesdej: X = np.array(["a","a","b"])
y = np.array(["x","y",np.nan], dtype=object)

encoder.fit(X,y)
encoder.mapping_["b"])
Gives key error 'b'
The early return on contingency_table.shape[0] == 1 bypasses the logic that handles categories with always-missing targets (x_such_that_y_is_always_missing). This results in some categories being missing from mapping_, which can lead to errors later in transform.

src/synthpop/data_processing/encoders.py
```diff
@@ -2,33 +2,168 @@
...
+            self.mapping_ = {contingency_table.index[0]: np.zeros(1)}
```
anneliesdej: Constant feature check might be better to make on rank(C)=0 so it is more in line with the functional description

src/synthpop/data_processing/encoders.py
```diff
@@ -37,11 +172,44 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
+        unique_values_in_x = np.unique([str(v) for v in X if not pd.isna(v)])
+        
+
+        if np.isin(unique_values_in_x,list(self.mapping_.keys()),assume_unique=True,invert=True).all():
+            raise ValueError("new values not seen during fitting when encoding.")
+
+        #if len(np.setdiff1d(unique_values_in_x,list(self.mapping_.keys()),assume_unique=True)) !=0:
+        #    raise ValueError("new values not seen during fitting when encoding.")
+
+        return np.array(
```
anneliesdej: X = np.array([np.nan, np.nan])
encoder = PCAEncoder()
encoder.mapping_ = {"a": [1.0, 2.0]}
encoder.n_features_out_ = 2
out = encoder.transform(X)
array([0., 0.])

You'd expect output to be array([[np.nan, np.nan.],
       [np.nan, np.nan]])

anneliesdej (changes requested):

src/synthpop/data_processing/encoders.py
anneliesdej: See line 177         if pd.isna(X).all():
            return np.zeros(X.shape[0])

src/synthpop/data_processing/encoders.py
```diff
@@ -174,18 +175,21 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
         # In that case, X cannot be sorted.
         # many routines of numpy for finding differences between lists depend on the items being sortable.
-        unique_values_in_x = np.unique([str(v) for v in X if not pd.isna(v)])
+        
         
 
-        if np.isin(unique_values_in_x,list(self.mapping_.keys()),assume_unique=True,invert=True).all():
-            raise ValueError("new values not seen during fitting when encoding.")
+        if hasattr(self,"mapping_"):
+            unique_values_in_x = np.unique([v for v in X if not pd.isna(v)])
```
anneliesdej: I just found out that in the case of the mixed incomparable types np.unique could potentially crash because it internally sorts the array. Safer would be to use set like:
unique_values_in_x = set(v for v in X if not pd.isna(v))
and then something like if not unique_values_in_x.issubset(self.mapping_.keys()):

src/synthpop/data_processing/encoders.py
anneliesdej: you should use pd.isna vectorised and not as a list (like pd.isna(X))

anneliesdej: here you should also use it vectorised as pd.isna(y_val)

anneliesdej: there is one space too many here :)

src/synthpop/data_processing/encoders.py
```diff
@@ -132,10 +132,10 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
         
         contingency_table = pd.crosstab(X_val,y_val,dropna=False).loc[x_such_that_y_is_not_always_missing]#the result of pd.crosstab is a pandas dataframe
 
+        x_such_that_y_is_always_missing = missing_contingency_table[missing_contingency_table[False]==0].index
```
anneliesdej, line 139: X = np.array(["a", "a","b","b","c"])
y = np.array(["x", "x","x","x","x"])
encoder = PCAEncoder
encoder.fit(X, y)
This gives a KeyError on False

anneliesdej (changes requested):

src/synthpop/data_processing/encoders.py
anneliesdej: In the test case of "1" and 1 it does not crash

anneliesdej (approved):

Issue #45, opened 2026-02-26
support for numpy arrays


KoomenErina:
This package assumes pandas dataframes as datatype for tables. Scikit-learn [writes ](https://scikit-learn.org/stable/faq.html#does-scikit-learn-work-natively-with-various-types-of-dataframes) that it has "limited support" for dataframes. It seems like numpy arrays are the default datatype for scikit-learn. This could mean that users that already know scikit learn might expect our package to work with numpy arrayas. Supporting numpy arrays might help with being compatible with scikit learn too.

KoomenErina:
Differences between pandas dataframe and numpy arrays:
|pandas dataframe| numpy array|
|--------------------|--------------|
|[hetrogeneous datatype ](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html)(mixing different types of columns)| [homogeneous datatype ](https://numpy.org/doc/stable/glossary.html#term-homogeneous)(one datatype for entire array)|
|2D| any number of dimensions|
|`df['column_name']` works by default| `data['field_name']` [works only](https://numpy.org/doc/stable/user/basics.indexing.html#field-access) for structured arrays|
|[has data manipulation methods (aggregating, grouping, sql-like operations) and statistics methods](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html)| [has mostly](https://numpy.org/doc/stable/reference/routines.array-manipulation.html) primitive matrix routines|
| [not fully optimized ](https://towardsdatascience.com/speed-testing-pandas-vs-numpy-ffbf80070ee7/) for preformance| [more](https://towardsdatascience.com/speed-testing-pandas-vs-numpy-ffbf80070ee7/) optimized. |

anneliesdej:
Below a list for pandas functions used in the PoC (potentially not complete). With models I mean methods:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=nl style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>


Pandas function   (pandas as pd) | Used where in the   PoC
-- | --
pd.api.types.is_numeric_dtype | Plots, metric,   synth
pd.concat | Models, encoder,   synth
[...]

PR #49, opened 2026-03-02
update dataflowdiagram datatypes

anneliesdej:
part of issue #45

### DIFF
--- a/docs/source/developer/dataflowdiagram.md
+++ b/docs/source/developer/dataflowdiagram.md
@@ -5,12 +5,12 @@ flowchart TD
       x: pd.DataFrame
       y=None|SF["Synthesiser.fit(x)"]-->
       i1{{Loops through x to pick column y each time}}-->CMF["CartMethod.fit(X: pd.DataFrame, y: pd.Series)"]-->icm{{Transform pd.DataFrame and pd.Series to np.Array}}-->
-      TCM["TreeClassifierMethod.fit(X: np.Array(string/float), y: np.Array(string))"]-->i2{{"Loops through arrays X to find where X is np.Array(string). Sends this to the encoder."}}-->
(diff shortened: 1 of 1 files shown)


docs/source/developer/dataflowdiagram.md
```diff
@@ -1,33 +1,31 @@
...
 flowchart TD
       U(["User"])-->|Step 1| S["Synthesiser()"]
       U-->|Step 2
       x: pd.DataFrame
       y=None|SF["Synthesiser.fit(x)"]-->
-      i1{{Loops through x to pick column y each time}}-->CMF["CartMethod.fit(X: pd.DataFrame, y: pd.Series)"]-->
-      TCM["TreeClassifierMethod.fit(X: pd.DataFrame, y: pd.Series)"]-->i2{{Loops through pd.DataFrame X to find categorical X. Sends this as pd.Series to the encoder.}}-->
-      PCAF["PCAEncoder.fit(X: pd.Series, y: pd.Series)"]--> PCAT["PCAEncoder.transform(X: pd.Series)"]-->|"X: pd.DataFrame (encoded)"|TCM---->DTC["DecisionTreeClassifier.fit(X (encoded), y(missing handled))
+      i1{{Loops through x to pick column y each time}}-->CMF["CartMethod.fit(X: pd.DataFrame, y: pd.Series)"]-->icm{{Transform pd.DataFrame and pd.Series to np.Array}}-->
+      TCM["TreeClassifierMethod.fit(X: np.Array(string/float), y: np.Array(string))"]-->i2{{"Loops through arrays X to find where X is np.Array(string). Sends this to the encoder."}}-->
```
KoomenErina: What do you mean with `np.Array(string/float)`? It seems like a heterogeneous datatype. Perhaps a dictionairy is needed here?

anneliesdej: Before encoding we have arrays of strings and arrays of floats. So here, X would be an array of strings or an array of floats. I could change it to np.Array(string)/np.Array(float) if that makes it more clear

KoomenErina (approved):

PR #50, opened 2026-03-03
Proposal for 45

KoomenErina:
changed typehints and docs to reflect solution of #45

### DIFF
--- a/docs/source/developer/code_standards_and_norms.md
+++ b/docs/source/developer/code_standards_and_norms.md
@@ -15,6 +15,8 @@ See this document as a tie-breaker when there are multiple equally well options,
 - [SOLID](https://en.wikipedia.org/wiki/SOLID)
 - It should always be possible to generate synthetic data within 5 lines of code. 
 - All classes that inherit from BaseEstimator should follow [the developer guides of scikit-learn](https://scikit-learn.org/stable/developers/develop.html)
+- Use numpy arrays where possible. 
+- The user of this package should be able to use pandas
 
 
 ## Standard topics in testing
--- a/docs/source/developer/code_standards_and_norms.md
+++ b/docs/source/developer/code_standards_and_norms.md
@@ -18,7 +18,8 @@ See this document as a tie-breaker when there are multiple equally well options,
 - Use numpy arrays where possible. 
 - The user of this package should be able to use pandas
 
-A synthesis method should accept pandas dataframes/series and heterogenous data.
+A synthesis method should accept pandas dataframes/series and heterogeneous data.
+Making the data homogeneous should happen within the synthesis method.
 It is best to use numpy within a synthesis method. See [#45](https://github.com/Synthpop-data/synthpop-py/issues/45).
 
 
--- a/docs/source/developer/code_standards_and_norms.md
+++ b/docs/source/developer/code_standards_and_norms.md
@@ -18,6 +18,11 @@ See this document as a tie-breaker when there are multiple equally well options,
 - Use numpy arrays where possible. 
 - The user of this package should be able to use pandas
 
+A synthesis method should accept pandas dataframes/series and heterogenous data.
+It is best to use numpy within a synthesis method. See [#45](https://github.com/Synthpop-data/synthpop-py/issues/45).
+
+
+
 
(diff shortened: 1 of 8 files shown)


src/synthpop/methods/base_synth.py
```diff
@@ -47,7 +47,7 @@ def fit(self,X:pd.DataFrame | None, y: pd.Series) -> Self:
         There should be an implementation of missing values support in case of missing values in feature columns.
         There should be a binary classifier to predict missing values if the target variable includes missing values.
         
-        :param X: Dataset of features
+        :param X: Dataset of features, may be heterogenous
```
anneliesdej: typo heterogeneous

src/synthpop/methods/base_synth.py
```diff
@@ -68,7 +68,7 @@ def transform(self, X: pd.DataFrame| None) -> pd.DataFrame:
 
         Calling `transform` before `fit` raises an error.
 
-        :param X: Input dataset
+        :param X: Input dataset, may be heterogenous
```
anneliesdej: same typo heterogeneous

docs/source/developer/code_standards_and_norms.md
```diff
@@ -15,6 +15,13 @@ See this document as a tie-breaker when there are multiple equally well options,
 - [SOLID](https://en.wikipedia.org/wiki/SOLID)
 - It should always be possible to generate synthetic data within 5 lines of code. 
 - All classes that inherit from BaseEstimator should follow [the developer guides of scikit-learn](https://scikit-learn.org/stable/developers/develop.html)
+- Use numpy arrays where possible. 
+- The user of this package should be able to use pandas
+
+A synthesis method should accept pandas dataframes/series and heterogenous data.
```
anneliesdej: typo heterogeneous

docs/source/developer/code_standards_and_norms.md
```diff
@@ -15,6 +15,13 @@ See this document as a tie-breaker when there are multiple equally well options,
...
+It is best to use numpy within a synthesis method. See [#45](https://github.com/Synthpop-data/synthpop-py/issues/45).
```
anneliesdej, line 23: maybe add that because of this the data within a synthesis method should be made homogeneous (with encoding for example)

src/synthpop/methods/cart_synth.py
anneliesdej, line 157: should this also include that it can be heterogeneous or is that overkill? Same for transform

src/synthpop/data_processing/encoders.py
```diff
@@ -30,7 +31,7 @@ def fit(self,X:pd.Series, y: pd.Series) -> Self:
         """
         return self
 
-    def transform(self,X:pd.Series) -> pd.DataFrame:
+    def transform(self,X:npt.ArrayLike) -> npt.ArrayLike:
```
anneliesdej: would the output not specifically be a numpy array of floats?  This would then also apply to the mean encoder

src/synthpop/data_processing/encoders.py
```diff
@@ -19,7 +20,7 @@ class PCAEncoder(TransformerMixin, BaseEstimator):
     def __init__(self, PCA_threshold:int = 30, explained_variance:float = 0.95):
         pass
 
-    def fit(self,X:pd.Series, y: pd.Series) -> Self:
+    def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
```
anneliesdej, line 23: would y here not be an array of strings? same goes for the mean encoder where it would be an array of floats right?

src/synthpop/methods/cart_synth.py
KoomenErina: that is implied by the datatype

src/synthpop/data_processing/encoders.py
KoomenErina: I am not quite sure if restricting to an array of strings is a good thing to do. Categorical data is quite often represented by an interger, as in Pandas. The code will not break if it is something else than a string.

anneliesdej (approved):

PR #53, opened 2026-03-06
48 update mean encoder with new data types

anneliesdej:


tests/unit/test_mean_encoder.py
```diff
@@ -4,142 +4,114 @@
...
+    assert result.mapping_ == expected_mapping
+
+def test_fit_calculates_means_fractional(fractional_data):
+    X, y = fractional_data
+    enc = MeanEncoder()
+    result = enc.fit(X, y)
+    expected_mapping = {"a": 1.5, "b": 2.5, "c": 3.5}
+    assert result is enc, "Fit should return self"
+    assert enc.n_features_in_ == 1, "self.n_features_in_ must equal 1 after fitting"
+    assert result.mapping_ == expected_mapping, "Encoder calculates incorrect mean values with fractional values"
```
KoomenErina: This (`test_fit_calculates_means` and `test_fit_calculates_means_fractional`)  seems like the same test but with different data. Perhaps this duplication can be avoided with parametrized tests? https://docs.pytest.org/en/stable/how-to/parametrize.html

src/synthpop/data_processing/encoders.py
```diff
@@ -54,25 +68,48 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
         :param y: Target column.
 
         Examples
```
KoomenErina, line 83: These examples don't render well in html. If you copy and paste from a terminal running python, it does render correctly

src/synthpop/data_processing/encoders.py
```diff
@@ -54,25 +68,48 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
-        # self.n_features_in_ = 1
 
-        # # Raises exception if y is not numeric
-        # if not pd.api.types.is_numeric_dtype(y):
-        #     raise TypeError(f"Column '{y.name}' must be numeric, got {y.dtype}")
+        # Required sklearn attributes
+        X, y = validate_data(X, y, ensure_2d=False, ensure_min_samples=1, y_numeric=True)
+
+        self.n_features_in_ = 1
+        self.feature_names_in_ = np.array(["x"], dtype=object)
```
KoomenErina: The feature name should come from the name of the pd.Series if `X` is a pd.series. Otherwise it does not need to be set.

src/synthpop/data_processing/encoders.py
```diff
@@ -54,25 +68,48 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
+
+        # Input validation
+        #if X.ndim != 1:
+        #    raise ValueError("X must be a 1D array.")
+        #if y.ndim != 1:
+        #    raise ValueError("y must be a 1D array.")
+        #if X.shape[0] != y.shape[0]:
+        #    raise ValueError("X and y must have the same length.")
+        if not np.issubdtype(y.dtype, np.number):
+            raise TypeError(f"Target column y must be numeric, got {y.dtype}.")
```
KoomenErina: is this needed in combination with `validate_data`?

tests/unit/test_mean_encoder.py
```diff
@@ -1,143 +1,165 @@
...
+    assert "1" in enc.mapping_, "fit does not handle mixed types in the feature" #makes all string
+
+# ----- transform test cases -----
+def test_transforms_uses_mapping():
     #Given a fitted estimator
     encoder = MeanEncoder()
-    encoder.mapping_ = {"red":1,"blue":2,"green":None}
-    encoder.feature_names_in_ = np.array(["color"])
+    encoder.mapping_ = {"red": 1, "blue": 2, "green": None}
+    encoder.feature_names_in_ = np.array(["colour"])
```
KoomenErina: feature names in is not needed when you don't use the output api

tests/unit/test_mean_encoder.py
```diff
@@ -1,143 +1,165 @@
...
+    X = np.array(["a", "b", "c"])
+    y = np.array([5, 5, 5])
+    enc = MeanEncoder().fit(X, y)
+    X_transformed = enc.transform(X)
+    assert np.all(X_transformed == 5), "Constant target should lead to constant encoding"
+
+def test_transform_raises_for_unseen_categories():
+    X = np.array(["a", "a", "b", "b", "c"])
+    y = np.array([1, 0, 2, 0, 3])
+    enc = MeanEncoder().fit(X, y)
```
KoomenErina: Since this test is about `transform`, fit should not be called. Instead, just set `mapping_`.  You don't need `X` and `y` either.

tests/unit/test_mean_encoder.py
```diff
@@ -1,143 +1,165 @@
...
+    X_new = np.array(["a", "b", "z"])  # 'z' is unseen
     with pytest.raises(ValueError):
-        encoder.transform(Z)    
+        enc.transform(X_new)
 
+def test_transform_wrong_shape():
+    X = np.array(["a", "a", "b", "b", "c"])
+    y = np.array([1, 0, 2, 0, 3])
+    enc = MeanEncoder().fit(X, y)
```
KoomenErina: `X`,`y`, and `fit` are not needed here.

tests/unit/test_mean_encoder.py
```diff
@@ -1,143 +1,165 @@
...
+    with pytest.raises(ValueError):
+        enc.transform(np.array([["a"], ["b"]]))  # 2D instead of 1D
+
+# ----- get_feature_names_out test cases -----
+def test_get_feature_names_out():
+    X = np.array(["a", "a", "b", "b", "c"])
+    y = np.array([1, 0, 2, 0, 3])
+    enc = MeanEncoder().fit(X, y)
+    names = enc.get_feature_names_out()
```
KoomenErina, line 152: `get_feature_names_out` can be called with an argument, see [here](https://scikit-learn.org/stable/glossary.html#term-get_feature_names_out) and [here](https://scikit-learn.org/stable/developers/develop.html#developer-api-for-set-output)

I think it is best if the name of the column being encoded (when present) is used in the name of the output column. This migh help to avoid name collisions.

For[ example](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.FunctionTransformer.html#sklearn.preprocessing.FunctionTransformer.get_feature_names_out), in the `FunctionTransformer` of sklearn is only defined if `feature_names_in_` is set.

src/synthpop/data_processing/encoders.py
```diff
@@ -82,28 +115,48 @@ def transform(self,X:npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optim
...
-        y = pd.Series([1, 0, 2, 0, 3], name='score')
+        >>> X = np.array(["a", "a", "b", "b", "c"])
+        >>> y = np.array([1, 0, 2, 0, 3])
 
-        encoder = MeanEncoder()
-        encoder.fit(X, y)
-        X_transformed = encoder.transform(X)
+        >>> encoder = MeanEncoder()
+        >>> encoder.fit(X, y)
+        >>> X_transformed = encoder.transform(X)
```
KoomenErina: you could include the numeric output in the examples

src/synthpop/data_processing/encoders.py
```diff
@@ -73,38 +73,39 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
 
         >>> encoder = MeanEncoder()
         >>> encoder.fit(X, y)
+        array([0.5, 0.5, 1.,  1.,  3. ], dtype=float32)
```
KoomenErina: this is not the output of `encoder.f(X,y)`

src/synthpop/data_processing/encoders.py
```diff
@@ -73,38 +73,39 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
         """
 
         # Required sklearn attributes
-        if isinstance(X,pd.Series):
-            self.feature_names_in_ = [X.name]
-        self.n_features_in_ = 1
-
         y = np.array([np.nan if (v is pd.NA or v is None) else v for v in y], dtype=float) #for pd.NA compatibility
```
KoomenErina: can be replaced by `pd.isna`

tests/unit/test_mean_encoder.py
KoomenErina: setting `feature_names_in_` is not needed here.

KoomenErina (changes requested):

src/synthpop/data_processing/encoders.py
```diff
@@ -84,7 +96,7 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
              dict(ensure_2d=False, ensure_min_samples=1, dtype='numeric', ensure_all_finite="allow-nan")
         ))
 
-        if isinstance(X,pd.Series) and X.name is not None:
+        if isinstance(X,pd.Series) and ~pd.isna(X.name):
```
KoomenErina: this gives a warning: ` : DeprecationWarning: Bitwise inversion '~' on bool is deprecated and will be removed in Python 3.16. This returns the bitwise inversion of the underlying int object and is usually not what you expect from negating a bool. Use the 'not' operator for boolean negation or ~int(x) if you really want the bitwise inversion of the underlying int.
    if isinstance(X,pd.Series) and ~pd.isna(X.name):`

KoomenErina (approved):

PR #61, opened 2026-03-19
55 implement replacenonewithvalue

KoomenErina:
replaces missing values with a value


src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,60 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+    >>> x_res,y_res = replace_missing.prepare_data_for_fit(X,y)
+    >>> x_res
+    array(['a', 'b', 'c', 'c'], dtype='<U1')
+    >>> y_res
+    array(['x', 'y', 'N.a.N.', 'z'], dtype='<U6')
+    >>> replace_missing.post_synth_transform(x_res,y_res)
+    array(['x', 'y', None, 'z'], dtype=object)
     """
+
+    def __init__(self,missing_marker= "N.a.N."):
```
anneliesdej: missing_marker is a mutable default argument. It shoud be str | None = None and then set in prepare_data_for_fit

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,60 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+        super().__init__()
+        self.missing_replacement = missing_marker
     
     def prepare_data_for_fit(self,X:npt.ArrayLike, y:npt.ArrayLike)-> tuple[npt.ArrayLike,npt.ArrayLike]:
         """
         Replaces missing values in the target with "N.a.N."
```
anneliesdej, line 109: The functional description of CART mentions both "N.a.N" and "N.a.N.". That should be changed

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,60 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
     Replace missing values by a specified value, and remove after synthesis.
+
+    :param missing_marker: The value to replace missing values with.
+
+    Examples
+    ========
+    >>> import numpy as np
+    >>> from synthpop.data_processing.missing_value_handling import ReplaceNoneWithValue
+    >>> X = np.array(["a","b","c","c"])
+    >>> y = np.array(["x","y",None,"z"])
```
anneliesdej, line 81: Only mentions None, when the detectoin is broader. Maybe also mention pd.NA and np.nan as possibles

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,60 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+        :return: a tuple ``(X,y)``. Leaves ``X`` unchanged. Replaces missing values in the target with "N.a.N.". Makes a copy of ``y``. 
         """
-        
-        return(pd.DataFrame(),pd.Series())
+
+        missing_mask = pd.isna(y)
+        if self.missing_replacement in y[~missing_mask]:
+            raise ValueError(f"the value {self.missing_replacement} already occurs in y")
+
+        result = copy.copy(y)
```
anneliesdej: It might be better to use numpy to create a copy of the array. You don't need to import copy then. You could use np.array(y, copy=True) for example

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,60 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
         Replaces "N.a.N." with missing values.
+
         :param X: the features for the target.
         :param y: the target
 
         :return:  The synthesised target with missing values.
         """ 
-        return pd.Series()
\ No newline at end of file
+        mask = y == self.missing_replacement
```
anneliesdej: Safer option: mask = np.equal(y, self.missing_replacement)

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,60 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+        if not mask.any():
+            return y
```
anneliesdej: Do we want y to be the same dtype here as the expected output?

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,60 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+    array(['x', 'y', 'N.a.N.', 'z'], dtype='<U6')
+    >>> replace_missing.post_synth_transform(x_res,y_res)
+    array(['x', 'y', None, 'z'], dtype=object)
     """
+
+    def __init__(self,missing_marker= "N.a.N."):
+        super().__init__()
+        self.missing_replacement = missing_marker
     
     def prepare_data_for_fit(self,X:npt.ArrayLike, y:npt.ArrayLike)-> tuple[npt.ArrayLike,npt.ArrayLike]:
```
anneliesdej, line 107: the type hint np.ArrayLike allows also for lists and padas Series, but the code assumes numpy-style indexing. It would be safer to transform the ArrayLike with np.asarray()

tests/unit/test_replace_missing_with_value.py
```diff
@@ -0,0 +1,77 @@
...
+    *[(np.array(["a","b","a"]),np.array(["x","y",missing_target],dtype=np.object_),np.array(["a","b","a"]),np.array(["x","y",missing_indicator],dtype=np.str_),missing_indicator) for missing_target in [None,pd.NA,np.nan] for missing_indicator in missing_indicators],
+    *[(np.array(["a","b",None],dtype=np.object_),np.array(["x","y","y"]),np.array(["a","b",None],dtype=np.object_),np.array(["x","y","y"]),"N.a.N.") ]
+    
+]
+@pytest.mark.parametrize("X_in,y_in,X_exp,y_exp,missing_indicator", get_test_data())
+def test_prepare_data_for_fit_numeric_correctness(X_in,y_in,X_exp,y_exp,missing_indicator):
+    replace_nan = ReplaceNoneWithValue(missing_marker = missing_indicator)
+
+    X_res,y_res = replace_nan.prepare_data_for_fit(X_in,y_in)
+    assert X_res is X_in
```
anneliesdej, line 28: Safer to test assert np.array_equal(X_res, X_in) as this enforces not just equality but also no copy of it

tests/unit/test_replace_missing_with_value.py
```diff
@@ -0,0 +1,77 @@
...
+
+    assert isinstance(y_res,pd.Series)
+    assert isinstance(X_res,pd.Series)
+
+    assert y_res.name == "someName_y"
+    assert X_res.name == "someName_X"
+
+def test_prepare_data_for_fit_error_when_nan_is_a_value():
+    X = np.array(["a","b"])
+    y = np.array(["x","N.a.N."])
```
anneliesdej: you should also add None to y. The functional description states you should get an error only if "N.a.N." already exists and there is a missing value

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,60 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
 class ReplaceNoneWithValue(BaseMissingValueHandler):
     """
     Replace missing values by a specified value, and remove after synthesis.
+
+    :param missing_marker: The value to replace missing values with.
```
anneliesdej, line 74: The functional description right now only states that "N.a.N" is seen as a special marker. Not sure if we should explicitly change the FD to mention that it can also be a different defined marker

tests/unit/test_replace_missing_with_value.py
```diff
@@ -0,0 +1,77 @@
+import numpy as np
+import pandas as pd
+import pytest
+from synthpop.data_processing.missing_value_handling import ReplaceNoneWithValue
+
+
+def get_test_data():
```
anneliesdej, line 7: I would like to also see a test with all values missing in y

tests/unit/test_replace_missing_with_value.py
```diff
@@ -0,0 +1,77 @@
...
+
+@pytest.mark.parametrize("X_in,y_in,X_exp,y_exp,missing_indicator", get_test_data())
+def test_prepare_data_for_fit_numeric_correctness_pandas(X_in,y_in,X_exp,y_exp,missing_indicator):
+    replace_nan = ReplaceNoneWithValue(missing_marker = missing_indicator)
+
+    X_res,y_res = replace_nan.prepare_data_for_fit(pd.Series(X_in,name="someName_X"),pd.Series(y_in,name="someName_y"))
+    assert np.array_equal(X_exp,X_res)
+    assert np.array_equal(y_exp,y_res)
+
+    assert isinstance(y_res,pd.Series)
```
anneliesdej, line 53: I would've expected this test to fail because result is stated as: return(X,result.astype(np.str_)). So I expected a numpy ndarray instead of pd.Series

anneliesdej (changes requested):

src/synthpop/data_processing/missing_value_handling.py
KoomenErina: yes

src/synthpop/data_processing/missing_value_handling.py
KoomenErina: strings in python are not mutable, see https://docs.python.org/3/reference/datamodel.html#immutable-sequences

```
>>> def test(x="abc"):
...     x[0] = "y"
...     return x
... 
>>> test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    import platform
    ^^^^^^
  File "<stdin>", line 2, in test
    import sys
     ^^^^
TypeError: 'str' object does not support item assignment
```

src/synthpop/data_processing/missing_value_handling.py
KoomenErina: changed in functional description

src/synthpop/data_processing/missing_value_handling.py
KoomenErina: The code is test with both `np.array` and `pd.Series`, and works

tests/unit/test_replace_missing_with_value.py
KoomenErina: `is` is only true if 2 symbols refer to the same object in memory. The examples from the docs of numpy seem to indicate that `np.array_equals` is simular to `==`

tests/unit/test_replace_missing_with_value.py
KoomenErina: `pd.Series` have a `astype` method too. See https://pandas.pydata.org/docs/reference/api/pandas.Series.astype.html

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,69 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
         :param y: the target.
 
-        :return: a tuple (X,y). Leaves X unchanged. Replaces missing values in the target with "N.a.N."
+        :return: a tuple ``(X,y)``. Leaves ``X`` unchanged. Replaces missing values in the target with "N.a.N.". Makes a copy of ``y``. 
         """
-        
-        return(pd.DataFrame(),pd.Series())
+        y_arr = self._copy_y(y)
+        missing_mask = pd.isna(y_arr)
+        if self.missing_replacement in y_arr[~missing_mask] and missing_mask.any():
```
anneliesdej: `in` can differ in behaviour. Especially with pd.Series as it checks for index not value:
`"a" in pd.Series(["a", "b"])`
`False`
For arrays it does not work vectorised but more like any(x == element for element in array). np.equal + np.any is preferred as it is vectorised and consistent across ndarray and Series (using \_\_array__)
->`np.any(np.equal(y_arr[~missing_mask], self.missing_replacement))`

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,69 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+            raise ValueError(f"the value {self.missing_replacement} already occurs in y")
+
+        # The result of np.copy is always a numpy array. If y is a pandas series, the expected output is a pandas series.
+        # So if y is a pandas series (or not numpy array), it is better to use copy.copy.
```
anneliesdej: This is not true. copy.copy gives a shallow copy on pd.Series: https://pandas.pydata.org/docs/reference/api/pandas.Series.copy.html

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,69 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+    >>> y_res
+    array(['x', 'y', 'N.a.N.', 'z'], dtype='<U6')
+    >>> replace_missing.post_synth_transform(x_res,y_res)
+    array(['x', 'y', None, 'z'], dtype=object)
     """
-    
+
+    def __init__(self,missing_marker:str = "N.a.N."):
+        super().__init__()
+        self.missing_replacement = missing_marker
```
anneliesdej, line 94: For readability it is nice to see an empty line before and after a function

tests/unit/test_replace_missing_with_value.py
```diff
@@ -0,0 +1,122 @@
...
+@pytest.mark.parametrize("X_in,y_in,y_exp,missing_indicator", get_test_data())
+def test_prepare_data_for_fit_numeric_correctness_pandas(X_in,y_in,y_exp,missing_indicator):
+    replace_nan = ReplaceNoneWithValue(missing_marker = missing_indicator)
+
+    x_pd = pd.Series(X_in,name="someName_X")
+    X_res,y_res = replace_nan.prepare_data_for_fit(x_pd,pd.Series(y_in,name="someName_y"))
+    assert X_res is x_pd
+    assert np.array_equal(y_exp,y_res)
+
+    assert isinstance(y_res,pd.Series)
```
anneliesdej, line 53: With pandas you also expect y_res.index == input.index which should be added to the test (but passes when I tested it)

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,69 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+
         :param X: the features for the target. 
         :param y: the target.
 
-        :return: a tuple (X,y). Leaves X unchanged. Replaces missing values in the target with "N.a.N."
+        :return: a tuple ``(X,y)``. Leaves ``X`` unchanged. Replaces missing values in the target with "N.a.N.". Makes a copy of ``y``. 
         """
-        
-        return(pd.DataFrame(),pd.Series())
+        y_arr = self._copy_y(y)
```
anneliesdej, line 116: Right now there is no handling or testing for empty data (like y = np.array([]). This error is given right now when y is empty (and X is not): '        # If y is a numpy array, it is faster to use np.copy.
>       y_arr[missing_mask] = self.missing_replacement
        ^^^^^^^^^^^^^^^^^^^
E       ValueError: could not convert string to float: 'N.a.N.'

src\synthpop\data_processing\missing_value_handling.py:119: ValueError'

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -68,25 +70,69 @@ def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
...
+            raise ValueError(f"the value {self.missing_replacement} already occurs in y")
+
+        # The result of np.copy is always a numpy array. If y is a pandas series, the expected output is a pandas series.
+        # So if y is a pandas series (or not numpy array), it is better to use copy.copy.
+        # If y is a numpy array, it is faster to use np.copy.
+        y_arr[missing_mask] = self.missing_replacement
+        return(X,y_arr.astype(np.str_))
 
     def post_synth_transform(self,X:npt.ArrayLike,y:npt.ArrayLike) -> npt.ArrayLike:
         """
```
anneliesdej, line 126: Here is also no handling or testing for empty y, with this error given: src\synthpop\data_processing\missing_value_handling.py:132: UFuncTypeError

anneliesdej (changes requested):

src/synthpop/data_processing/missing_value_handling.py
KoomenErina: a "deep copy" by pd.Series.copy  is not the same as a deep copy by `copy.deepcopy`.  There were no problems with the shallow copy being made in this code, but that migth be because of copy on write behaviour. It now uses `pd.series.copy(deep=True)` to be independen of c.o.w behaviour.

src/synthpop/data_processing/missing_value_handling.py
anneliesdej: Comment is outdated now as code is explicitly made to handle array, series and lists

anneliesdej (approved):

PR #62, opened 2026-03-20
57 reusable code to sample from decisiontree

anneliesdej:


src/synthpop/methods/tree_utils.py
```diff
@@ -1,14 +1,15 @@
...
 import numpy.typing as npt
 from sklearn.tree import BaseDecisionTree
+from sklearn.base import BaseEstimator
 from sklearn.utils.validation import check_is_fitted
+from sklearn.utils import check_random_state
 
 T = TypeVar("T")
 
-class LeafNodeSampler:
+class LeafNodeSampler(BaseEstimator):
```
KoomenErina: why inherit from `BaseEstimator`?

src/synthpop/methods/tree_utils.py
```diff
@@ -67,9 +68,39 @@ def fit_sampler(self, tree: BaseDecisionTree, X: npt.ArrayLike, y: npt.ArrayLike
...
+        leaf_ids = tree.apply(X)
+        self._leaf_map = {}
+
+        for leaf_id, target in zip(leaf_ids, y):
+            if leaf_id not in self._leaf_map:
+                self._leaf_map[leaf_id] = {}
+
+            leaf_hist = self._leaf_map[leaf_id]
+
+            if pd.isna(target):
```
KoomenErina, line 99: The target should not contain missing values. If it does, it is a bug.

src/synthpop/methods/tree_utils.py
```diff
@@ -87,4 +118,36 @@ def sample_from_leaves(self, X_syn: npt.ArrayLike) -> np.ndarray:
...
+            
+            leaf_hist = self._leaf_map[leaf_id]
+            values = np.array(list(leaf_hist.keys()), dtype=object)
+            counts = np.array(list(leaf_hist.values()), dtype=np.int32)
+
+            if counts.sum() == 0:
+                y_syn[i] = np.nan
+                continue
+
+            probs = counts / counts.sum()
```
KoomenErina: Does this introduce floating point problems?

tests/unit/test_leaf_node_sampler.py
```diff
@@ -0,0 +1,312 @@
...
+        ([[0], [1], [2]], [0, 1, 1], [10, 10, 20],
+            {10: {0: 1, 1: 1}, 20: {1: 1}}),
+
+        # --- pandas Series ---
+        (pd.DataFrame([0, 1, 2]).values, pd.Series([0, 1, 1]), [10, 10, 20],
+            {10: {0: 1, 1: 1}, 20: {1: 1}}),
+
+        # --- strings ---
+        (np.array(["a", "b", "c"]), np.array(["x", "z", "y"]), np.array([1, 1, 2]),
+            {1: {"x": 1, "z": 1}, 2: {"y": 1}}),
```
KoomenErina, line 25: I think it should raise an error when `X` is not numeric.  The decisiontrees cannot handle non-numeric input. Encodings should happen before `LeafNodeSampler` is used

tests/unit/test_leaf_node_sampler.py
```diff
@@ -0,0 +1,312 @@
...
+
+        (np.array([1, 2, 3]), np.array(["a", "a", "b"]), np.array([1, 1, 2]),
+            {1: {"a": 2}, 2: {"b": 1}}),
+
+        # --- mixed: None, np.nan, pd.NA ---
+        (np.array([0, 1, 2, 3]), np.array([None, np.nan, pd.NA, 1], dtype=object), np.array([10, 10, 20, 20]),
```
KoomenErina: dealing with missing is not needed here. If there are missing values in the target, an error should be raised.

tests/unit/test_leaf_node_sampler.py
```diff
@@ -0,0 +1,312 @@
...
+            {10: {np.nan: 2}, 20: {np.nan: 1, 1: 1,}}),
+        
+        # --- mixed string and integer ---
+        (np.array([1, "1", 1], dtype=object), np.array([1, "1", 1.0], dtype=object), [1, 2, 1],
+            {1: {1: 2}, 2: {"1": 1}}),
```
KoomenErina, line 36: this input should throw an error

KoomenErina (changes requested):
This class is used in the end of the data process. Here we should not accept "dirty data".

src/synthpop/methods/tree_utils.py
anneliesdej: so sklearn's clone works which is needed within the tree classes

src/synthpop/methods/tree_utils.py
anneliesdej: We concluded that the code with the if statement would be removed. Any missing values will just be treated as keys. Because of the CART design, at this point there should not be missing values anymore, so a warning will be stated. However, we don't throw an error as this function is a standalone and not dependent on CART

tests/unit/test_leaf_node_sampler.py
anneliesdej: We agreed to keep the function a standalone so any ArrayLike input should just be accepted

tests/unit/test_leaf_node_sampler.py
anneliesdej: we agreed to handle this the same as in the encoders. The input should be accepted as this is an issue of the tree methods and not this function

src/synthpop/methods/tree_utils.py
anneliesdej: added a clone function in the class so it can be used as sampler.clone()

src/synthpop/methods/tree_utils.py
anneliesdej: changed the rng to avoid this

tests/unit/test_leaf_node_sampler.py
KoomenErina: `np.random.RandomState` is deprecated, see https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState

It gives an error with test discovery for me.

When I change it to `np.random.default_rng(random_state)`, the test discovery does work, but multiple tests fail because the code in tree_utils seems to depend on those deprecated methods, such as `randint()` (see https://numpy.org/doc/stable/reference/random/generated/numpy.random.RandomState.randint.html)

src/synthpop/methods/tree_utils.py
```diff
@@ -45,27 +52,34 @@ def transform(self, X_syn):
...
+        for sampling.
 
-        :param X: Feature matrix used to determine leaf membership
-        :param y: Target values corresponding to the rows of `X`
-        :return: Fitted sampler
-        
-        Example:
-        --------
-        
+        This function can be used with missing values, but those are not expected at this point in the synthesis process
```
KoomenErina: Writing "This function can be used with missing values" is true, but might let people think that it would work out of the box. Perhaps something like "If `tree` supports missing values in `y`, then this method works with missing values. Or perhaps something like "This method passes any missing or non-missing values of `y` to `tree`."

KoomenErina (changes requested):

tests/unit/test_leaf_node_sampler.py
KoomenErina: This margin is very wide for a unit test. Furthermore, there might be a seed for which this tests fails. The test might fail if fewer samples are taken. 
I would recommend using a stub for the random number generator. This would have the following benefits:
- make explicit what behaviour is assumed/expected from the random number generator
- The test becomes independent of the statistical/random properties (sample size, seed). The fact that this test passes is partly due to "luck". 
- the possibility to test edgecases. (if the random number generator has a perfect uniform output, is the output of sample_from_leaves uniform as well? or does the `side` parameter introduce a bias?

src/synthpop/methods/tree_utils.py
KoomenErina: why the explicit dtypes? all tests pass when you remove them. We don't specifically require that the counts are int64.

KoomenErina, line 160: if  `total == 0`, then the leaf node was empty after fitting. That means that the decion tree has no path that leads to this leafnode.  So this node will never be in the output of `apply`. 

The only scenarios in which `total == 0` would be true are:
- X_syn is empty
- `_leaf_map` has been corrupted

if `X_syn` is empty, then either `apply` would raise an error, or the for loop would be skipped.

if `_leaf_map` has been corrupted, then outputting `np.nan` is hiding a serious bug. This might cause a situation where a variable is never missing in the observed data, but is sometimes missing in the synthetic data. Finding the reason for that would be very difficult. 

Besides, setting some values of `y_syn` to `np.nan` assumes that `y_syn` is numeric. 

I would suggest either raising an error if  `total == 0` or removing this check and let the error happen by `random_state_.intergers`.

KoomenErina: is there a way to vectorize this? That might make a significant performance impact

tests/unit/test_leaf_node_sampler.py
KoomenErina: an empty histogram is not something that should happen. see my comment in tree_utils.py line 162

KoomenErina: an empty histogram is not something that should happen. see my comment in tree_utils.py line 162.

KoomenErina (changes requested):

tests/unit/test_leaf_node_sampler.py
```diff
@@ -0,0 +1,244 @@
...
+    leaf_ids=[10, 10, 10, 10, 10]
+    sampler = helper_make_sampler(
+        leaf_map={10: {np.nan: 2, pd.NA: 1, None: 1, 1: 1}},
+        leaf_ids=leaf_ids
+    )
+
+    y_syn = sampler.sample_from_leaves(leaf_ids)
+
+    assert len(y_syn) == 5
+    assert any(pd.isna(v) for v in y_syn)
```
KoomenErina: using `any` leaves room for bugs. For example, it might happen that only one `None` is included, but `np.nan` and `pd.na` will never be in the output. I think a test that verifies that all 3 of the missings can ben in the output. Perhaps parametrize the 3 missings?

src/synthpop/methods/tree_utils.py
```diff
@@ -33,58 +34,150 @@ def __init(tree_sampler: LeafNodeSampler | None = None):
...
+        n_samples = len(leaf_ids)
+
+        if leaf_ids.ndim != 1:
+            raise ValueError(f"leaf_ids must be 1-dimensional with shape (n_samples,), got shape {leaf_ids.shape} instead.")
+        if n_samples == 0:
+            raise ValueError(f"leaf_ids must be non-empty.")
+
+        y_syn = np.empty(n_samples, dtype=object)
+
+        unique_leaves, inverse_indices = np.unique(leaf_ids, return_inverse=True)
```
KoomenErina, line 143: why `return_inverse=True`?

KoomenErina (changes requested):

src/synthpop/methods/tree_utils.py
anneliesdej: Was part of an optimalisation experiment

tests/unit/test_leaf_node_sampler.py
```diff
@@ -0,0 +1,249 @@
...
+    sampler._leaf_map = {10: {0: 3, 1: 1}}
+    sampler.random_state_ = rng
+
+    leaf_ids = [10, 10, 10, 10]
+
+    y = sampler.sample_from_leaves(leaf_ids)
+
+    assert list(y) == [0, 1, 0, 0]
+
+import pytest
```
KoomenErina:
```suggestion

```

tests/unit/test_leaf_node_sampler.py
```diff
@@ -0,0 +1,249 @@
...
+import numpy as np
```
KoomenErina:
```suggestion
```

tests/unit/test_leaf_node_sampler.py
```diff
@@ -0,0 +1,249 @@
...
+import pandas as pd
```
KoomenErina:
```suggestion
```

KoomenErina (changes requested):

KoomenErina (approved):

PR #74, opened 2026-03-27
implement missingvaluepredictor and tests

anneliesdej:


tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,249 @@
...
+
+    model._all_missing = False
+    model._no_missing = False
+
+    X = {"a": [1, 2, 3, 4], "b": [10, 20, 30, 40]}
+    y = np.array([100, 200, 300, 400])
+    out = model.post_synth_transform(X, y)
+
+    assert len(out) == 4
+    assert np.isnan(out[1])
```
KoomenErina: It is not obvious why the second entry should be `nan`.

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,249 @@
...
+        self.transform_called = True
+        return np.array([1.0 if x is not None else 0.0 for x in X])
+
+class DummySampler:
+    def __init__(self):
+        self.fit_called = False
+        self.sample_called = False
+        pass
+
+    def fit_sampler(self, leaf_ids, z):
```
KoomenErina: It looks like there is no test to validate that `fit_sampler` is called with the correct arguments. I see test that check if `fit_sampler` is called, but `last_leaf_ids` is never used.

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,249 @@
...
+        self._leaf_map = True
+        self.random_state_ = True
+        self.fit_called = True
```
KoomenErina: if `last_z` and `last_leaf_ids` are set, it is already implied that `fit_sampler` has been called

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,249 @@
+import numpy as np
+import pandas as pd
+import pytest 
+from sklearn.tree import BaseDecisionTree
+from sklearn.base import BaseEstimator
+
+from synthpop.data_processing.missing_value_handling import MissingValuePredictor
+# ----- fixtures -----
```
KoomenErina: this seems like a good usecase for pytest fixtures (https://docs.pytest.org/en/stable/explanation/fixtures.html)

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -16,56 +18,198 @@ class BaseMissingValueHandler(metaclass=ABCMeta):
...
         """
+        # input validation
+        X_val, y_val = self._validate_X_y_dict(X, y)
+
+        self.tree_ = clone(self.tree) if self.tree else DecisionTreeClassifier(min_samples_leaf= 5)
+        self.tree_sampler_ = self.tree_sampler.clone() if self.tree_sampler else LeafNodeSampler()
         
-        return(pd.DataFrame(),pd.Series())
+        # implementation
+        z = pd.isna(y_val).astype(int)
```
KoomenErina: why cast to `int`? and shouldn't it be `np.int_`?

if z is boolean, you don't need `z==1`, and in line 210 you won't need `z_sampled == 1'

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -16,56 +18,198 @@ class BaseMissingValueHandler(metaclass=ABCMeta):
...
+                encoded = encoder.transform(values).astype(np.float32)
+            
+            X_encoded.append(encoded.reshape(-1, 1))
+
+        X_matrix = np.hstack(encoded.reshape(-1, 1))
+
+        leaf_ids = self.tree_.apply(X_matrix)
+        z_sampled = self.tree_sampler_.sample_from_leaves(leaf_ids).astype(int)
+
+        y_out = y.astype(object).copy()
```
KoomenErina: if we specialize `MissingValuePredictor` to only deal with numeric `y`, then casting to `object` is not needed. Object arrays have less optimal preformance compared to float arrays. And if you use `np.nan`, then you don't need the object dtype

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -16,56 +18,198 @@ class BaseMissingValueHandler(metaclass=ABCMeta):
...
+        y_out[z_sampled == 1] = np.nan
```
KoomenErina: if `z_sampled` is boolean, you don't need `==1`. That might be slightly better for preformance?

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -16,56 +18,198 @@ class BaseMissingValueHandler(metaclass=ABCMeta):
...
+                encoded = values.astype(np.float32)
+            else:
+                encoded = encoder.transform(values).astype(np.float32)
+            
+            X_encoded.append(encoded.reshape(-1, 1))
+
+        X_matrix = np.hstack(encoded.reshape(-1, 1))
+
+        leaf_ids = self.tree_.apply(X_matrix)
+        z_sampled = self.tree_sampler_.sample_from_leaves(leaf_ids).astype(int)
```
KoomenErina: I don't think you need to cast to int, boolean would work better

KoomenErina (changes requested):
The test are not checking enough. If line 151 of missing_value_handling.py is replaced with `self.tree_.fit(None, z)`, all tests still pass. Same goes for calls to the encoder. 

At some points you only test if the function is called at all, but not if the function is called with the correct arguments. 

Furthermore, the missing value predictor needs some integration tests.

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -119,6 +118,25 @@ def _validate_X_y_dict(cls, X: Dict[str, npt.ArrayLike], y: npt.ArrayLike) -> tu
...
+            self.feature_order_ = feature_order
+
+        X_encoded = []
+        for col in feature_order:
+            values = np.asarray(X[col])
+            encoder = self.encoders_.get(col, None)
+            if encoder is None:
+                encoded = values.astype(np.float32)
+            else:
+                encoded = encoder.transform(values).astype(np.float32)
```
KoomenErina: casting to `np.float32` should not happen here.

KoomenErina:
There is a general approach to unit testing a class that has dependencies that I would recommend. 
The assumption of that approach is that methods of the dependencies are called because either you need the output (as mostly in this package) or because of the side effects (for example, output to the user, write to a file/database, not really applicable in this package). To isolate the unit that is being tested, you can use stubs/mocks. The general layout (independent of sklearn) of the test is as follows:

The code to be tested:
```python
class ClassToTest():
   def __init__(self, some_dependency = SomeDependency()):
        self._some_dependency = some_dependency

  def method_being_tested(self,input_value):
       input_processed = input_value +10
       intermediate_step = self._some_depencency.some_method(input_processed, some_param = 123)
       final_result = intermediate_stop*2
       return final_result
```

The code of the stub:
```python
class SomeDependencyStub
  def __init__(self,some_method_return_value):
       self.some_method_return_value = some_method_return_value

  def some_method(self,input_data,some_param):
       self.some_method_input_data = input_data
       self.some_method_some_param = some_param
       return self.some_method_return_value
``` 

The code of the unit test would be as follows:

```python
def test_method_being_tested_applies_some_method():
  some_dependency =  SomeDependencyStub(some_method_return_value= 43)
  being_tested = ClassToTest(some_depencency = some_dependency )

  result = being_tested.method_being_tested(11)

 # Assert that some_depencency.some_method is called correctly
  assert some_dependency.some_method_input_data == 11 + 10
  assert some_dependency.some_method_some_param == 42

 # assert that the return value of some_depencency.some_method is being used correctly
  assert result == 43*2
```

By setting the return values of the stub and asserting how it is called, you write down what you expect the real depdency to return when given certain inputs. In other words, when you stub like this, you make assumptions about dependencies. If the unit test of `method_being_tested` pass, but the integration tests do not pass, then there is likely a mismatch between assumptions and reality. 

In this way, you can add test for every different return value of the dependency: empties, different datatypes.

This does not test how often the dependencies are called, or in what order. That would require more advanced mocking.

This is a general strategy. In reality tests can look different. Requirements by sklearn alter this setup slightly. Using pytest fixtures and parametrization can change how the code looks, but still does the same in essence.

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+
+        def fit(self, X, y):
+            self.fitted = True
+            self.mapping_ = True
+            self.fit_called = True
+            return self
+    
+        def transform(self, X):
+            self.transform_called = True
+            return np.array([1.0 if x is not None else 0.0 for x in X])
```
KoomenErina: you need to test more than one possible output of the encoder.

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+            self.X = None
+            self.y = None
+            pass
+
+        def fit(self, X, y):
+            self.X = X
+            self.y = y
+            self.fit_called = True
+            return self
+
```
KoomenErina, line 65: you need to test more possible outputs

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+    "X, y",
+    [
+        ({"a": [1, 2], "b": [3, 4]}, [1, 2]),
+        ({"a": np.array([1, 2]), "b": pd.Series([3, 4])}, np.array(["a", "b"])),
+        ({"a": pd.Series(["a", "b"]), "b": np.array([3, 4])}, pd.Series([1, 2])),
+        ({1: [1, 2], 2: [3, 4]}, [None, pd.NA]),
+        ({"a": [1, "1"], "b": [1, "a"]}, np.array([1, "1"], dtype=object))
+    ]
+)
+def test_validate_accepts_valid_X_y(X, y):
```
KoomenErina: unit tests for non-public methods should be avoided. 
non-public methods are called by the public methods. 
Test the expected behaviour/input-output of the public methods instead.

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+    assert np.array_equal(sampler.last_z, expected_z)
+
+def test_prepare_data_no_missing_data_flow(predictor):
+    X = {"cat": ["a", "b", "c", "d"], "num": [1, 2, 3, 4]}
+    y = [10, 20, 30, 40] 
+
+    X_out, y_out = predictor.prepare_data_for_fit(X, y)
+
+    # core branch behaviour
+    assert predictor._no_missing
```
KoomenErina: test public methods and properties instead of internal properties

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+    assert not predictor._all_missing
+
+    assert predictor.tree_ is None
+    assert predictor.tree_sampler_ is None
```
KoomenErina: why should `tree_` and `tree_sampler_` be  `None`? The trailing underscore suggest it are learned parameters, but if they are `None`, what has been learned?

This comment applies to other test with assertions that these parameters are  `None` too.

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+
+    # encoding still happens
+    assert "cat" in predictor.encoders_
+    assert predictor.encoders_["cat"].fit_called
+    assert predictor.encoders_["cat"].transform_called
```
KoomenErina: It should be tested that the methods are called with the correct arguments, not just that they are called

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+
+    assert predictor.encoders_["num"] is None
```
KoomenErina: if  `encoders_["num"]` should be `None`, why is it in the dictionary? can it be left out?

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+    for k in X:
+        assert np.array_equal(X_out[k], X[k])
+    assert np.array_equal(y_out, y)
+
+def test_prepare_data_encoding_called_only_to_categorical(predictor):
+    X = {"cat": ["a", "b", "c"], "num": [1, 2, 3]}
+    y = [0, 1, 0]
+    predictor.prepare_data_for_fit(X, y)
+    encoder = predictor.encoders_["cat"]
+    assert "cat" in predictor.encoders_
```
KoomenErina: if  `encoder = predictor.encoders_["cat"]` does not raise an exception, then ` assert "cat" in predictor.encoders_` is not needed

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,428 @@
...
+    assert encoder.transform_called
+    assert predictor.encoders_["num"] is None
+
+def test_prepare_data_tree_and_sampler_called(predictor):
+    X = {"cat": ["a", "b", "c", "d"]}
+    y = [0, np.nan, 1, 0]
+    predictor.prepare_data_for_fit(X, y)
+    assert predictor.tree_.fit_called
+    assert hasattr(predictor.tree_sampler_, "last_z")
+    assert hasattr(predictor.tree_sampler_, "last_leaf_ids")
```
KoomenErina: assert the value of `last_z` and `last_leaf_ids`, not just exsistance.

KoomenErina (changes requested):

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,352 @@
...
+def test_prepare_data_missing_data_flow_correct(predictor):
+    predictor.encoding.transform_return = np.array([1, 1, 1, 1])
+    predictor.tree.apply_return = np.array([0, 1, 2, 3])
+
+    X = {"cat": ["a", "b", "a", "b"]}
+    y = [0, None, 0, 1]
+
+    X_out, y_out = predictor.prepare_data_for_fit(X, y)
+
+    assert np.array_equal(predictor.encoders_["cat"].fit_inputs, (X["cat"], pd.isna(y))), "encoding input should be original X and and missingness mask"
```
KoomenErina:
```suggestion
    assert np.array_equal(predictor.encoders_["cat"].fit_inputs, (X["cat"], np.array([False,True,False,False]))), "encoding input should be original X and and missingness mask"
```

It is better to use explicit test values instead of using an external library method. This way, it is clear that the mask is a numpy boolean array and not a list of integeres, for example

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,352 @@
...
+
+    tree = predictor.tree_
+    tree_X, z = tree.fit_inputs
+    
+    assert tree_X.shape == (4, 1) #Tree expects matrix
+    assert np.array_equal(tree_X, predictor.encoding.transform_return.reshape(-1, 1)), "X input of the tree should be the output of the encoding but reshaped to 2D"
+
+    sampler = predictor.tree_sampler_
+    leaf_ids, z_sampler = sampler.fit_inputs
```
KoomenErina: rename `leaf_ids`,`z_sampler` to something like `given_sampler_input_leaf_ids/y_values` so that it is clear that these leaf_ids represent the "actual" part of the test

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,352 @@
...
+    assert np.array_equal(y_out, [0, 0, 1])
+
+def test_prepare_data_no_missing_data_flow(predictor):
+    predictor.encoding.transform_return = np.array([1, 1, 1, 1])
+    X = {"cat": ["a", "b", "c", "d"], "num": [1, 2, 3, 4]}
+    y = [10, 20, 30, 40] 
+
+    X_out, y_out = predictor.prepare_data_for_fit(X, y)
+
+    assert predictor.tree_ is None, "no tree should be build when there are no missing values"
```
KoomenErina:
```suggestion
    assert predictor.tree_.fit_inputs is None, "no tree should be build when there are no missing values"
```

testing if `self.tree_` is `None` is not an effective strategy to test if no tree has been build. It only tests if the `else`  (on line 176 missing_value_handling.py) block has been activated. 

That `else` block should be removed. Setting a public property to `None` means that users that want to use `tree_` need to both check if the object has a"tree_" property and check that it is not `None`.

This goes for the other places where it is tested that a dependency should be None as well

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,352 @@
...
+    assert predictor.tree_sampler_ is None, "no sampler should be used when there are no missing values"
+
+    enc_num = predictor.encoders_["num"]
+    assert enc_num is None
```
KoomenErina:
```suggestion
    assert len(predictor.encoders_) ==1
```
No dictionary keys should be set for numeric columns. Making assertions about the not None keys and asserting the lenght proves that the not-None keys are all keys

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,352 @@
...
+
+    enc_cat = predictor.encoders_["cat"]
+    assert enc_cat is not None
```
KoomenErina:
```suggestion
```
`enc_cat` is used on the next line, that proves it is not None

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,352 @@
...
+    assert np.array_equal(fit_y, pd.isna(y)), "encoding input should be original missingness mask"
+
+    for col in X:
+        assert np.array_equal(X_out[col], np.array(X[col]))
+
+    assert np.array_equal(y_out, np.array(y)), "nothing should change to output data if no missing"
+    assert np.array_equal(list(X_out.keys()), predictor.feature_order_)
+
+def test_prepare_data_for_fit_mixed_types(predictor):
+    predictor.encoding.transform_return = pd.array([1, 1, 1, 1, 1, 1]) #without the code fails
```
KoomenErina: if you set the return value to this in the fixture, you don't need to add this line to all test that are not about the encoding

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,352 @@
...
+    predictor._all_missing = False
+    predictor._no_missing = False
+
+    X = {"a": [1, 2, 3, 4], "b": [10, 20, 30, 40]}
+    y = np.array([100, 200, 300, 400])
+    predictor.feature_order_ = list(X.keys())
+
+    out = predictor.post_synth_transform(X, y)
+
+    assert predictor.tree_sampler_.sample_inputs is not None, "Sampler should be called"
```
KoomenErina:
```suggestion
    assert np.array_equals(predictor.tree_sampler_.sample_inputs, predictor.tree_.apply_return) "Sampler should be called"
```

test if the function is called with the correct arguments, not just if an argument has been given

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,352 @@
...
+
+def test_post_synth_transform_dataflow(predictor, stub_tree, stub_sampler):
+    predictor.tree_ = stub_tree
+    predictor.tree_.apply_return = np.array([0, 1, 2, 3])
+    predictor.tree_sampler_ = stub_sampler
+    predictor.tree_sampler_.sample_return = np.array([False, True, False, False])
+    predictor._all_missing = False
+    predictor._no_missing = False
+    predictor.encoders_ = {
+        "a": type("E", (), {"transform_return": np.array([1, 2, 3, 4]),
```
KoomenErina: Use the encoder stub that has already been writen.

tests/unit/test_leaf_node_sampler.py
```diff
@@ -124,6 +125,7 @@ def test_sample_from_leaves_various_input_types(leaf_ids):
             assert val in expected_values
 
     assert len(y_syn) == len(leaf_ids)
+    assert y_syn.dtype == sampler._y_dtype
```
KoomenErina, line 145: I see no test of `fit_sampler` that sets `_y_dtype`.
There should be a test that the dtype is set correctly when the input is a string array or boolean array or numeric array.

KoomenErina (changes requested):

tests/unit/test_missing_value_predictor.py
```diff
@@ -7,79 +7,72 @@
...
 @pytest.fixture
-def dummy_sampler():
-    class _DummySampler:
-        def __init__(self):
-            self.fit_called = False
-            self.sample_called = False
-            self.last_leaf_ids = None
-            self.last_z = None
+def stub_sampler():
+    class SamplerStub:
```
KoomenErina, line 34: why a class definition within a function?

tests/unit/test_missing_value_predictor.py
```diff
@@ -121,86 +112,74 @@ def test_validate_raises_invalid_y(y):
...
+    assert predictor.feature_order_ == ["a", "b"], "Feature order contract should be handled correctly"
+    assert np.array_equal(predictor.feature_order_, list(X_out.keys())), "output feature order should be the same as input"
 
-# ----- prepare data for fit tests -----
 def test_prepare_data_missing_data_flow_correct(predictor):
+    predictor.encoding.transform_return = np.array([2, 2, 2, 2])
+    predictor.tree.apply_return = np.array([0, 1, 2, 3])
+
     X = {"cat": ["a", "b", "a", "b"]}
     y = [0, None, 0, 1]
```
KoomenErina: Testing for other "types of missing" (np.nan, pd.NA) are still missing. This can be quickly done by parametrizin this test

tests/unit/test_missing_value_predictor.py
```diff
@@ -121,86 +112,74 @@ def test_validate_raises_invalid_y(y):
...
-    # encoding still happens
-    assert "cat" in predictor.encoders_
-    assert predictor.encoders_["cat"].fit_called
-    assert predictor.encoders_["cat"].transform_called
-
-    assert predictor.encoders_["num"] is None
+    enc_cat = predictor.encoders_["cat"]
+    fit_X, fit_y = enc_cat.fit_inputs
+    assert np.array_equal(fit_X, X["cat"]), "encoding input should be original X"
+    assert np.array_equal(fit_y, pd.isna(y)), "encoding input should be original missingness mask"
```
KoomenErina:
```suggestion
assert np.array_equal(fit_y, [False]*len(y))
```

no need to use pandas here

tests/unit/test_missing_value_predictor.py
```diff
@@ -121,86 +112,74 @@ def test_validate_raises_invalid_y(y):
...
 
-    # core branch behaviour
-    assert predictor._no_missing
-    assert not predictor._all_missing
+    assert predictor.tree_ .fit_inputs is None, "no tree should be build when there are no missing values"
+    assert predictor.tree_sampler_.fit_inputs is None, "no sampler should be used when there are no missing values"
 
-    assert predictor.tree_ is None
-    assert predictor.tree_sampler_ is None
+    assert "num" not in predictor.encoders_
```
KoomenErina:
```suggestion
```
you already prove that "cat" is the only key in the dictionary

tests/unit/test_missing_value_predictor.py
```diff
@@ -266,151 +220,105 @@ def test_prepare_data_does_not_mutate_inputs(predictor):
...
 
-    expected_mask = np.array([False, True, False, True]) #from DummySampler
+    assert np.array_equal(predictor.tree_sampler_.sample_inputs, predictor.tree_.apply_return), "Sampler should be called"
     assert out.shape == y.shape
-    assert np.array_equal(np.isnan(out), expected_mask)
+    assert np.array_equal(np.isnan(out), predictor.tree_sampler_.sample_return)
 
 
-def test_post_synth_all_missing(predictor, dummy_tree, dummy_sampler):
+def test_post_synth_all_missing(predictor, stub_tree, stub_sampler):
```
KoomenErina:
```suggestion
def test_post_synth_all_missing(predictor)
```

stub_tree and stub_sampler are not used

tests/unit/test_missing_value_predictor.py
```diff
@@ -266,151 +220,105 @@ def test_prepare_data_does_not_mutate_inputs(predictor):
...
-
     X = {"a": [1, 2, 3]}
     y = np.array([1, 2, 3])
-    predictor.feature_order_ = list(X.keys())
     out = predictor.post_synth_transform(X, y)
 
     assert np.all(np.isnan(out))
 
-def test_post_synth_no_missing(predictor, dummy_tree, dummy_sampler):
+def test_post_synth_no_missing(predictor, stub_tree, stub_sampler):
```
KoomenErina:
```suggestion
def test_post_synth_all_missing(predictor)
```

tests/unit/test_missing_value_predictor.py
```diff
@@ -266,151 +220,105 @@ def test_prepare_data_does_not_mutate_inputs(predictor):
...
+def test_post_synth_transform_dataflow(predictor, stub_tree, stub_sampler, stub_encoder):
+    predictor.tree_ = stub_tree
+    predictor.tree_.apply_return = np.array([0, 1, 2, 3])
+    predictor.tree_sampler_ = stub_sampler
+    predictor.tree_sampler_.sample_return = np.array([False, True, False, False])
     predictor._all_missing = False
     predictor._no_missing = False
+    encoder_a = stub_encoder
+    encoder_a.transform_return = np.array([1, 2, 3, 4])
+    encoder_b = type(stub_encoder)()  # new instance
```
KoomenErina: if the class definition of the encoder stub is not in a function, you don't need to use `type(stub_encoder)`. You can use copy or create a `EncoderStub` directly

tests/unit/test_missing_value_predictor.py
```diff
@@ -266,151 +220,105 @@ def test_prepare_data_does_not_mutate_inputs(predictor):
     assert X == X_copy
 
     # ----- post synth transform tests -----
-def test_post_synth_transform_basic(predictor, dummy_tree, dummy_sampler):
-    predictor.tree_ = dummy_tree
-    predictor.tree_sampler_ = dummy_sampler
-    predictor.encoders_ = {"a": None, "b": None}
-
+def test_post_synth_transform_basic(predictor, stub_tree, stub_sampler):
```
KoomenErina: Does this test check anything that is not tested in `test_post_synth_transform_dataflow`? if not, this test can be removed

tests/unit/test_leaf_node_sampler.py
KoomenErina: why the dataframe, and `.values`? would a numpy array suffice?

KoomenErina: should the `_seed` of `sampler2` be set?

src/synthpop/methods/tree_utils.py
```diff
@@ -51,11 +51,14 @@ def __init__(self, random_state: int | np.random.Generator | None = None):
...
-                via `np.random.default_rng`.
-            - If `numpy.random.Generator`, it is used directly.
+                via `np.random.default_rng`. This generator is reset with each call so output
+                is consistent between calls.
+            - If `numpy.random.Generator`, it is used directly. This generator is not reset with each
+                call, so the state advances with each call.
             - If `None`, a default seed (42) is used to ensure reproducibility.
         """
         self.random_state = random_state
+        self._y_dtype = None
```
KoomenErina: `y_dtype` is inferred from data, and should be treated as a learned parameter. Setting it to `None` might cause confusion with `NoneType`.

KoomenErina (changes requested):

tests/unit/test_leaf_node_sampler.py
```diff
@@ -164,6 +164,7 @@ def test_sampling_deterministic_with_stub_rng():
     sampler = LeafNodeSampler()
     sampler._leaf_map = {10: {0: 3, 1: 1}}
     sampler.random_state_ = rng
+    sampler._y_dtype = None #for test purpose only
```
KoomenErina: this test starts with setting up `sampler` as if it has been fitted to a sequence of integers. `y_dtype` would be `np.int_` in this scenario.  Setting `_y_dtype` to `None` makes this test pass because the default dtype of `np.empty` is `np.float64`. 

I would recommend setting  `_y_dtype` to an integer dtype in this test.

src/synthpop/methods/tree_utils.py
```diff
@@ -58,7 +59,7 @@ def __init__(self, random_state: int | np.random.Generator | None = None):
             - If `None`, a default seed (42) is used to ensure reproducibility.
         """
         self.random_state = random_state
-        self._y_dtype = None
+        self._y_dtype = _UNSET
```
KoomenErina:
```suggestion
````
Learned parameters should not be set in the `__init__`

File: src/synthpop/methods/tree_utils.py [hunk 39]
```diff
@@ -5,6 +5,7 @@
 import warnings
 
 T = TypeVar("T")
+_UNSET = object()
```
KoomenErina:
```suggestion
```

src/synthpop/methods/tree_utils.py
```diff
@@ -5,6 +5,7 @@
 import warnings
 
 T = TypeVar("T")
```
KoomenErina:
```suggestion
```
typevar is not used anymore

tests/unit/test_missing_value_predictor.py
KoomenErina: should raise NotFittedError (https://scikit-learn.org/stable/modules/generated/sklearn.exceptions.NotFittedError.html#sklearn.exceptions.NotFittedError)

KoomenErina (changes requested):

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,319 @@
...
+    assert np.array_equal(predictor.feature_order_, list(X_out.keys())), "output feature order should be the same as input"
+
+@pytest.mark.parametrize(
+        "y", [[0, None, 0, 1], [0, np.nan, 0, 1], [0, pd.NA, 0, 1]],
+)
+def test_prepare_data_missing_data_flow_correct(predictor, y):
+    predictor.encoding.transform_return = np.array([2, 2, 2, 2])
+    predictor.tree.apply_return = np.array([0, 1, 2, 3])
+
+    X = {"cat": ["a", "b", "a", "b"]}
```
KoomenErina: This test does not test if numeric input in `X` is handled correctly. 
Consider expanding the current `X` to include multiple numeric columns and multiple categorical columns at the same time.

combining the numerical components of `X` with the encoded inputs is not trivial, and should be tested.

tests/unit/test_missing_value_predictor.py
```diff
@@ -0,0 +1,319 @@
...
+    predictor.tree_ = stub_tree
+    predictor.tree_.apply_return = np.array([0, 1, 2, 3])
+    predictor.tree_sampler_ = stub_sampler
+    predictor.tree_sampler_.sample_return = np.array([False, True, False, False])
+    predictor._all_missing = False
+    predictor._no_missing = False
+    predictor.encoders_ = {}
+    predictor.feature_order_ = ["a", "b"]
+
+    X = {
```
KoomenErina, line 315: add multiple categorical columns to `X`. This test now only checks if numeric data follows `feature_order_`.

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -2,70 +2,245 @@
...
+        z = 1 → missing
+        z = 0 → observed
+    
+    Then samples:
+        z ~ Bernoulli(P(z=1|x))
+
+    :param encoding: Default is a :doc: `MeanEncoder` <synthpop.data_processing.encoders.MeanEncoder>.
+    :param tree: Decision tree classifier. Default is [DecisionTreeClassifier(min_samples_leaf=5)](https://scikit-learn.org/stable/modules/generated/sklearn.tree.DecisionTreeClassifier.html)
+    :param tree_sampler: Leaf node sampler. Default is :py:meth:LeafNodeSampler
+        .
```
KoomenErina: add some example usage

KoomenErina (changes requested):

KoomenErina (approved):

PR #80, opened 2026-04-08
Updated developer docs information and structure

anneliesdej:


docs/source/developer/checklist_for_review.md
```diff
@@ -1,54 +1,119 @@
...
+### Correctness and design
+- [ ] Implementation correctly satisfies the specification
+- [ ] Code is not more complex than necessary
+- [ ] No duplicate code is unnecessarily introduced
+- [ ] Each component has a clear and single responsibility
+- [ ] SOLID principles are respected where applicable
+- [ ] Any edge case you see is covered by unit tests
+### Data and numerical integrity
+- [ ] Input data is not unintentionally mutated
+- [ ] No data leakage occurs between stages (fit/transform/generate)
```
KoomenErina: is this something that can happen by accident?

docs/source/developer/checklist_for_review.md
```diff
@@ -1,54 +1,119 @@
...
+- [ ] Numerical computations are stable (no unexpected NaNs or Infs)
```
KoomenErina: I don't think stable is the right word. Is this basically the same as that has been implemented correctly?

docs/source/developer/checklist_for_review.md
```diff
@@ -1,54 +1,119 @@
...
+- [ ] Floating point comparisons are handled safely
+### API consistency
+- [ ] Naming conventions are consistent with the codebase
+- [ ] Function signatures follow established patterns
```
KoomenErina: what do you mean with established patterns?

KoomenErina (changes requested):

docs/source/developer/checklist_for_review.md
anneliesdej: By mistake yes. If information is used that should not be available in this stage, if global states interfer, if fitted values are recomputed. I could add examples or remove this point from the checklist. What would your preference be?

docs/source/developer/checklist_for_review.md
anneliesdej: It overlaps with floating point issues, so this line will be removed

docs/source/developer/checklist_for_review.md
anneliesdej: This one is basically a summary of the other points, parameter naming, order, defaults, etc. I will clarify this

docs/source/developer/checklist_for_review.md
KoomenErina: I think that warning people of a danger is in general a good thing. But the design should usually specify which data to use in a process. Do you agree that this kind of data leakage is a design problem and not so much an implementation problem? I think that other kinds of leakage (unintentional memory acces, log messages, ) are more of an implementation issue (although unlikly in our context). 

I would like to see examples, but perhaps not in this document

KoomenErina (approved):

PR #94, opened 2026-04-20
implement sample method

### LINKED ISSUE #73: Implement Sample Method
Fill in the maquette of SampleMethod in sample_synth.py with the code. Write unit tests as well

anneliesdej:


src/synthpop/methods/sample_synth.py
```diff
@@ -3,30 +3,101 @@
...
+            or not hasattr(self, "n_samples_")
+            or not hasattr(self, "random_state_")):
+            raise NotFittedError("SampleMethod is not fitted. Call `fit` first.")
+        
+        n = len(X) if X is not None else self.n_samples_
+        
+        cum_counts = np.cumsum(self.counts_)
+        r = self.random_state_.integers(0, self.counts_.sum(), size=n)
+        idx = np.searchsorted(cum_counts, r, side="right")
+        sampled = self.values_[idx]
```
KoomenErina: This seems very simular to `LeafNodeSampler`. Is it possible to extract a common utility function? Or perhaps some other way of code reuse?

tests/unit/test_sample_method.py
```diff
@@ -0,0 +1,188 @@
...
+
+# ----- fit tests -----
+@pytest.mark.parametrize(
+    "y",
+    [
+        pd.Series([1, pd.NA, 3], name="my_target"),
+        pd.Series([1.1, 2.2, np.nan], name="my_target"),
+        pd.Series(["a", "b", "c"], name="my_target", dtype="category"),
+        pd.Series([None, False, True], name="my_target"),
+        pd.Series([0], name="my_target"),
```
KoomenErina: Every value of `y` results in an uniform probabilty distribution. Including a value for `y` that has a non-uniform distrubtion would make the test stronger.

tests/unit/test_sample_method.py
```diff
@@ -0,0 +1,188 @@
...
+    X = pd.DataFrame(index=range(50))
+
+    result1 = model1.transform(X)
+    result2 = model2.transform(X)
+
+    # Not guaranteed, but overwhelmingly likely
+    assert not result1.equals(result2)
+
+
+def test_transform_approximate_distribution():
```
KoomenErina: test more distributions, including extreme distributions (perfect uniform, one possible outcome).

KoomenErina (changes requested):

src/synthpop/methods/sample_synth.py
KoomenErina: The utility function can look something like this:
```python

def sample_array(random_number_generator, counts,values,n_samples):
   cum_counts = np.cumsum(counts)
   r = self.random_state_.integers(0, counts.sum(), size=n_samples)
   idx = np.searchsorted(cum_counts, r, side="right")
   sampled = values[idx]
   return sampled
```

from this method it could be called something like this:
```python
sampled = sample_array(self.random_state_, self.counts_,self.values_)
```

from `LeafNodeSampler.transform` it can be called something like this:
```python
...
sampled = sample_array(rng, counts, values)

for i, v in zip(indices, sampled):
   y_syn[i] = v
...
```

anneliesdej: Added issue #96 to also implement this change in the `LeafNodeSampler`

src/synthpop/methods/tree_utils.py
```diff
@@ -5,6 +5,21 @@
 import warnings
 from sklearn.exceptions import NotFittedError
 
+def sample_array(rng: np.random.default_rng, counts: np.ndarray, values: np.ndarray, n_samples: int):
```
KoomenErina:
```suggestion
def sample_array(rng: np.random.Generator, counts: np.ndarray, values: np.ndarray, n_samples: int):
```

`np.random.default_rng` is not a type, it's a function.

src/synthpop/methods/tree_utils.py
```diff
@@ -5,6 +5,21 @@
...
+    """
+    Helper function that draws samples with replacement from the empirical distribution of an array.
+
+    :param rng: a random number generator. Often `np.random.default_rng`.
```
KoomenErina:
```suggestion
    :param rng: a random number generator. Often `np.random.Generator`.
```

src/synthpop/methods/tree_utils.py
KoomenErina, line 192: is it possible to use `sample_array` here?

tests/unit/test_sample_method.py
```diff
@@ -0,0 +1,202 @@
...
+
+    expected_index = expected.index
+    model_index = pd.Index(model.values_)
+
+    expected_non_nan = set(v for v in expected_index if not pd.isna(v))
+    model_non_nan = set(v for v in model_index if not pd.isna(v))
+
+    assert model_non_nan == expected_non_nan
+
+    assert any(pd.isna(v) for v in model.values_) == any(pd.isna(v) for v in expected_index)
```
KoomenErina: I do not understand what this tries to assert. 
I read this as: "`model.values_` has at least one NA <=>  `expected_index` has at least one NA". 
If I understand correctly, the `==` compares two booleans in this statement. So it does not check if the NA is in the right place. If `model.values_` has multiple NAs and `expected` only one, the this assert still holds.

tests/unit/test_helper_sample_array.py
```diff
@@ -0,0 +1,149 @@
...
+            v = next(self.values)
+            return low + (v % (high - low))
+
+        out = []
+        for _ in range(size):
+            v = next(self.values)
+            out.append(low + (v % (high - low)))
+        return np.array(out)
+
+def test_sample_array_deterministic_with_stub_rng():
```
KoomenErina: The name of the test does not reflect what it is trying to test. 
Using stubs is a technique to test how the code being tested interacts with dependencies.
It seems like this test is about the process from random numbers to sampled values. Perhaps something like `test_sample_array_maps_random_to_values`?

There are 2 things that are usually tested when stubbing is used:
1. validate that the stubbed dependency is called with the correct data.
2. validate that the output of the dependency is used as intended. 

This test validates the second point, but not the first.

tests/unit/test_helper_sample_array.py
```diff
@@ -0,0 +1,149 @@
...
+    rng = np.random.default_rng(42)
+
+    counts = np.array(counts)
+    values = np.array(values, dtype=object)
+
+    out = sample_array(rng, counts, values, n_samples=1000)
+
+    if any(pd.isna(v) for v in values):
+        assert any(pd.isna(out))
+
```
KoomenErina, line 58: I see no test for arrays with `dtype=StringDType` with the appropiate missing value. You don't need to change every test, but at least one testcase should demonstrate that those arrays are supported.

tests/unit/test_helper_sample_array.py
```diff
@@ -0,0 +1,149 @@
...
+import pytest 
+
+from synthpop.methods.tree_utils import sample_array
+
+# ----- stubs -----
+class StubRNG:
+    def __init__(self, values):
+        self.values = iter(values)
+
+    def integers(self, low, high, size=None):
```
KoomenErina, line 13: A stub should not have this much logic. A stub should do nothing other than store the arguments it is called with and returns a "canned" result.  Adding logic to a stub has serveral disadvantages:
- more logic means more opportunity for bugs
- it obscures expected behaviour of the dependencies. It is not clear to me what behaviour is expected/what the assumption are of the RNG.
- it makes the unit test more difficult to understand.

tests/unit/test_helper_sample_array.py
```diff
@@ -0,0 +1,149 @@
...
+
+    assert out.shape == (0,)
+
+def test_sample_array_mismatched_lengths():
+    rng = np.random.default_rng(0)
+
+    counts = np.array([1, 2])
+    values = np.array([10])  # mismatch
+
+    with pytest.raises(IndexError):
```
KoomenErina: Why the choice for an `IndexError`? Would a `ValueError` be more appropiate?
An index error usually occurs when you try to acces an array out of bounds.

tests/unit/test_sample_method.py
```diff
@@ -0,0 +1,202 @@
...
+)
+def test_fit_stores_distribution_and_metadata(y):
+    model = SampleMethod().fit(None, y)
+
+    assert model.target_name_ == "my_target"
+    assert model.n_samples_ == len(y)
+
+    assert model.counts_.sum() == model.n_samples_
+
+    expected = pd.value_counts(y, dropna=False)
```
KoomenErina: This test uses the same logic as the implementation. This reduces the validity of the test.

Suppose a breaking change occurs in a new pandas version that changes `value_counts` (for example: "nan" v.s. pd.NA).
Then this test and the implementation could fail in the same way, resulting in a passing test while not correct.

This test does not show or explicitly validate how missings are handled. For example, it could convert all missings to strings and still pass.

tests/unit/test_sample_method.py
```diff
@@ -0,0 +1,202 @@
...
+
+    expected_index = expected.index
+    model_index = pd.Index(model.values_)
+
+    expected_non_nan = set(v for v in expected_index if not pd.isna(v))
+    model_non_nan = set(v for v in model_index if not pd.isna(v))
+
+    assert model_non_nan == expected_non_nan
```
KoomenErina: I do not understand what this is trying to assert. Maybe add a assert text?

KoomenErina (changes requested):

src/synthpop/methods/tree_utils.py
anneliesdej: Not in scope for this PR, will be worked on in Issue #96

tests/unit/test_helper_sample_array.py
anneliesdej: Not in scope for this PR. The helper is now focused on sample method en will be extended to the Leaf Node Sampler in issue #96

tests/unit/test_sample_method.py
anneliesdej: Added to tests

KoomenErina (approved):

PR #95, opened 2026-04-20
Proposal random numbers

KoomenErina:
**Current status**: proof of concept.

**How it works:**
The user is forced to suply a seed (`int`) when creating a `Synthesiser` object. This is the root seed.
This seed is used to derive seeds for the instances of classes where a RNG is needed (such as `LeafNodeSampler`). This means that each instance of `LeafNodeSampler` has it's  unique seed. Let's call this the "instance seed".

The classes that need random numbers call the classmethod `reproducibility.Reseed.get_rng` with the instance seed to get a random number generator. These RNGs are based on the root seed and the instance seed. If the root seed and instance seed are the same, then the sequence of number generated by the RNG will also be the same. 

To safely manage switching seeds, the `Reseed` class can be used as a [context manager](https://book.pythontips.com/en/latest/context_managers.html#implementing-a-context-manager-as-a-class). The root seed is changed to the seed given to `generate` when entering the `with` block. When `Reseed.get_rng` is called with an instance seed, the combination of rootseed and instance seed is different, so the resulting RNG will produce a different sequence of numbers.. The root seed is changed back when exiting the `with` block. 

See demo_repro.py line 46 - 69 for how this would look like for the user. See demo_repro.py line 0 - 46 for how this might be implemented. 

**Open questions:**
- how should the instance seeds look like? integers or `SeedSequences`? -> integers, for compatibility, and combining with root seed.
- How should seeds be combined? -> passing a list of integers to default_rng()
- If, for example, `CartMethod` is used independent of `Synthesiser`, then the user needs to call `Reseed.set_root_seed` or use in a `with rep.Reseed(seed)` block. This is possibly confusing. The cause of this is that we only force the user to supply a seed when creating a `Synthesiser` object. We could either use a hardcoded default, or use a non-reproducible default, or raise a clear exception reminding the user that it should supply a seed. -> the last option, raise an exception informing the user to set the seed.


anneliesdej (changes requested):
The new proposal seems good, but it would be better to see how it is implemented in, for example, the `LeafNodeSampler`. For the open questions, if you already have an idea, please work out pros and cons. We could also plan a meeting about this.

Furthermore, the PR needs cleaning up :)

src/synthpop/reproducibility.py
```diff
@@ -0,0 +1,83 @@
...
+            cls._seed_sequence = seed_sequence
+
+
+    @classmethod
+    def get_new_seed(cls) -> int:
+        """
+        Returns a seed that can be used to make a RNG.
+        The seed is based on the root seed.
+        """
+        return cls._seed_sequence.spawn(1)[0].generate_state(1)
```
anneliesdej, line 71: This returns a np array and not a plain int, while we kind of have a seed is int contract. It would be better to return a SeedSequence instead of an int. If we change fully to SeedSequence this could be replaced with a class method, for example:
```{python}
@classmethod
def spawn(cls, n: int = 1) -> list[SeedSequence]:
    """
    Spawn independent child SeedSequences from the root.
    """
    if cls._root_ss is None:
        raise RuntimeError("Root seed not set. Call Reseed.set_root_seed(...).")

    return cls._root_ss.spawn(n)
```

src/synthpop/reproducibility.py
```diff
@@ -0,0 +1,83 @@
...
+    The root seed is the basis for all random behaviour in this package.
+    The root seed is used to create a `numpy.random.SeedSequence`.   
+    """
+    _seed_sequence = None
+    """
+    The seed sequence is used to provide proper initialisation for all RNGs used in this package, even if the user provided seed is suboptimal.
+    """
+
+    @classmethod
+    def set_root_seed(cls,seed:int,seed_sequence = None):
```
anneliesdej, line 42: Better would be:
```{python}
@classmethod
def set_root_seed(cls, seed: int):
    """
    Initialise the root SeedSequence from a user-provided seed.
    """
    cls._root_ss = SeedSequence(seed)
```

This removes the need to manually "combine" seeds and we can just use SeedSequence.spawn. You would also need to remove both _reed_seed and _seed_sequence is None and just make _root_ss: SeedSequence | None = None

src/synthpop/reproducibility.py
```diff
@@ -0,0 +1,83 @@
...
+    def get_new_seed(cls) -> int:
+        """
+        Returns a seed that can be used to make a RNG.
+        The seed is based on the root seed.
+        """
+        return cls._seed_sequence.spawn(1)[0].generate_state(1)
+
+
+    @classmethod
+    def get_rng(cls,seed) ->np.random.Generator:
```
anneliesdej: By changing the set_root_seed you can also change get_rng ith something like this:
```{python}
@classmethod
def get_rng(cls, instance_ss: SeedSequence) -> np.random.Generator:
    """
    Create an RNG from a spawned SeedSequence.
    """
    if cls._root_ss is None:
        raise RuntimeError("Root seed not set. Call Reseed.set_root_seed(...).")

    return np.random.default_rng(instance_ss)
```

src/synthpop/reproducibility.py
anneliesdej: Although you would probably change this a little as now a new instance seed (sequence) would have to be called using Reseed.spawn(1)[0] which is less clear than Reseed.get_new_seed()

src/synthpop/synthesiser.py
```diff
@@ -11,15 +11,17 @@ class Synthesiser:
...
     :param first_column_method: The method for synthesising the first column. This is a special case, since there are no predictors available.
     """
-    def __init__(self, 
+    def __init__(self, random_seed: int,
                  column_order: list[str] | list[int] | None = None,
                  default_syn_method: BaseSynthMethod | None = None,
-                 special_syn_method: dict[str, BaseSynthMethod] | None = None) -> None:
+                 special_syn_method: dict[str, BaseSynthMethod] | None = None,
+                 ) -> None:
         pass
```
anneliesdej, line 25: You could make the implementation change more clear by putting the works in the __init__, like `Reseed.set_root_seed(seed)`

src/synthpop/reproducibility.py
```diff
@@ -0,0 +1,83 @@
+import numpy as np
+from numpy.random import SeedSequence
+
+
+
+class Reseed():
```
anneliesdej: SeedManager might be a more intuitive name or things like RandomStateManager

src/synthpop/reproducibility.py
anneliesdej: create_rng might be a better name

src/synthpop/reproducibility.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Manages random numbers and reproducibility in this package.
+
+    Instances of this class can be used as a context manager to temporary switch seed:
+    
+    Example:
+        >>> from reproducibility import Reseed
```
anneliesdej: It might be nice (design choice) to use it as `from reproducibility import Reseed as RNG`

docs/source/functional descriptions/reproducibility.md
```diff
@@ -1,32 +1,34 @@
...
-It might even be a formal requirement for some users. 
-Furthermore, a predictable package helps the user to build a reliable process to make a synthetic dataset.
-Random behaviour can frustrate debugging efforts. 
+Synthetic data generation requires both determinism and high-quality randomness. These goals are not contradictory but must be explicitly engineered. Reproducibility ensures that results are deterministic and auditable. Randomness quality ensures that generated data does not exhibit artificial structure or leak information about the source of the data.
 
-The methods to make a synthetic datasets fundamentally need randomness. 
-If the randomness is not random enough it might reduce the quality of the generated synthetic datasets by introducing patterns that are not in the observed data.
-It is not excluded that a lack of randomness can make it possible to reconstruct the observed data. 
+Poor handling of either undermines trust. Non-determinism breaks pipelines and debugging, while also makes it impossible to recreate the synthetic data. Weak randomness can introduce bias, or in extreme cases, enable reconstruction of the original data.
```
KoomenErina: I am not sure that it only happens in "extreme cases"

src/synthpop/reproducibility.py
KoomenErina: full caps as import alias is unusual in python. RNG stands for random number generator. The acronym is used often for instances of random number generators.

src/synthpop/reproducibility.py
KoomenErina: How does this use the root seed? Does this still allow for multiple synthetic datasets?

src/synthpop/reproducibility.py
KoomenErina: How would we combine the root seed(sequence)  with the instance seed(sequence)?

src/synthpop/reproducibility.py
KoomenErina: How does this remove the need to manually combine seeds?

src/synthpop/reproducibility.py
anneliesdej: We wouldn't manually combine root and instance seeds anymore, but rely on SeedSequence. The root spawns everything where no seperate instance seed concept is needed. So the root seed would initialise a SeedSequence and instance seeds are derived using .spawn().
```{python}
root_ss = SeedSequence(root_seed)
instance_ss = root_ss.spawn(1)[0]
rng = np.random.default_rng(instance_ss)
```
Here the root seed defines the experiment and each spawn is an independent component stream. The way I understand it, but correct me if I'm wrong, SeedSequence is based on hashing and an internal counter state so no runtime randomness is involved

src/synthpop/reproducibility.py
anneliesdej: We would not be "combining seeds" anymore in the sense of manually merging multiple entropy sources (e.g., `[root_seed, instance_seed]`). Instead, we construct a single root `SeedSequence` and derive all instance-specific seeds using `.spawn()`. So the manual part is only very small right now but could be completely gone

src/synthpop/reproducibility.py
KoomenErina: If the root seed is changed using the context manager (to generate another synthetic dataset), the instance seed sequences remain unaffected. Can the user still generate multiple independent synthetic datasets?

src/synthpop/reproducibility.py
anneliesdej: The root seed is used only to initialise a SeedSequence. All randomness in the system is deried from this object . Right now my example does not use root seed access in get_rng. The flow would be like root seed (int) -> SeedSequence(root) -> spawn() -> instance seedSequence -> default_rng(). But it can probably still be combined in the get_rng. In generate, the root seed would then need to be overrided in some way also using SeedSequence(override_seed).

docs/source/functional descriptions/reproducibility.md
anneliesdej: I will change it to some cases

src/synthpop/reproducibility.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+
+class RandomStateManager():
+    """
+    Manages random numbers and reproducibility in this package.
+
+    Instances of this class can be used as a context manager to temporary switch seed:
+    
+    Example:
```
anneliesdej:
```suggestion
    Examples
    --------
```
Coherence

src/synthpop/reproducibility.py
```diff
@@ -0,0 +1,95 @@
...
+    @classmethod
+    def create_new_seed(cls) -> int:
+        """
+        Returns a seed that can be used to make a RNG.
+        The seed is based on the root seed.
+        It is used to make instance seeds and seeds for external dependencies (legacy).
+        The reason that the instance seeds are integers is to facilitate combining the root seed and instance seed.
+
+        :returns: an integer that can be used as a seed.
+
```
anneliesdej, line 63:
```suggestion

    Examples
    --------
```
Coherence, but it is needed to check if indentation is correctly added from this suggestion

anneliesdej (approved):
2 coherence suggestions in documentation but else approved

PR #99, opened 2026-04-29
97 update to simplify encoders

### LINKED ISSUE #97: Update to simplify encoders
guidelines:
- input and output are 2D numpy arrays
- code duplication is resolved where applicable
-  dtype `StringDtype(na_object=np.nan)` is used as the default dtype for strings
- `np.nan` is the only representation of missingness. 
- consistent naming

This means that code (including tests) regarding pandas input/output or other representations of missingness should be removed.

KoomenErina:
closes #97 :

**Current status:**
- 2D input supported
- output is always 2D
- uses `StringDType`  and only ` np.nan` 
- unified shared/duplicate code.
- optimized
- consistent choices for empty arrays

### DIFF
--- a/docs/source/functional descriptions/PCA-encoding.md
+++ b/docs/source/functional descriptions/PCA-encoding.md
@@ -51,7 +51,7 @@ Only the first $k$ principal components are kept as part of the encoding.
 |missing | missing | missing|
 |any specific non-missing value| always missing for that specific feature value | missing|
 |any specific non-missing value| sometimes but not always missing for that specific feature value|the encoding treats the missing target value as a normal value and the default PCA is applied
-|many different non-missing values| the same constant over all values of the feature (any non-missing value)| The rotation matrix of PCA becomes an identity matrix. The contingency table becomes a 1D vector of the number of occurrences of the feature level. |
+|many different non-missing values| the same constant over all values of the feature (any non-missing value)| The rotation matrix of PCA becomes an identity matrix. The contingency table becomes a 1D vector of the number of occurrences of the feature level. The encoding is an scaled and centered version of count encoding |
 |constant (one non-missing value for all rows)| many different non-missing values or constant (any non-missing value) for that specific feature value| The contingency table becomes a constant vector. After centring, this vector is the zero vector. The feature should be encoded with 0.|
 
 ### 6.1 Missing values
--- a/docs/source/functional descriptions/PCA-encoding.md
+++ b/docs/source/functional descriptions/PCA-encoding.md
@@ -51,7 +51,7 @@ Only the first $k$ principal components are kept as part of the encoding.
 |missing | missing | missing|
 |any specific non-missing value| always missing for that specific feature value | missing|
 |any specific non-missing value| sometimes but not always missing for that specific feature value|the encoding treats the missing target value as a normal value and the default PCA is applied
-|many different non-missing values| the same constant over all values of the feature (any non-missing value)| The rotation matrix of PCA becomes an identity matrix. The contingency table becomes a 1D vector of the number of occurrences of the feature level. The encoding is an scaled and centered version of count encoding |
+|many different non-missing values| the same constant over all values of the feature (any non-missing value)| The rotation matrix of PCA becomes an identity matrix. The contingency table becomes a 1D vector of the number of occurrences of the feature level. The encoding is a scaled and centred version of count encoding |
 |constant (one non-missing value for all rows)| many different non-missing values or constant (any non-missing value) for that specific feature value| The contingency table becomes a constant vector. After centring, this vector is the zero vector. The feature should be encoded with 0.|
 
 ### 6.1 Missing values
--- a/src/synthpop/data_processing/encoders.py
+++ b/src/synthpop/data_processing/encoders.py
@@ -71,10 +71,12 @@ unseen = observed - seen
 
         unique_vals,rev_index = np.unique(X_val,return_inverse=True)
         #Note that rev_index does not reconstruct X_val when X_val is a stringDType array with np.nan:
-        # x_val = np.array(["a", "a", "b", "b", "c",np.nan, "c"],dtype = str_dtype)
-        # unique_vals,rev_index = np.unique(x_val,return_inverse=True)
-        # unique_vals[rev_index]
-        # array(['a', 'a', 'b', 'b', 'c', 'c', 'c'],dtype=StringDType(na_object=nan))
+        #>>> x_val = np.array(["a", "a", "b", "b", "c",np.nan, "c"],dtype = str_dtype)
+        #>>> unique_vals,rev_index = np.unique(x_val,return_inverse=True)
+        #>>> unique_vals[rev_index]
+        #>>> array(['a', 'a', 'b', 'b', 'c', 'c', 'c'],dtype=StringDType(na_object=nan))
+        #The only differences between the reconstruction and the original are the missing values.
+        # So to avoid searching X_val again to create a reverse index, we use this one instead and correct the missing values.
         
         mapped_vals = np.array([self.mapping_[v] for v in unique_vals],dtype=np.float32)
 
--- a/src/synthpop/data_processing/encoders.py
+++ b/src/synthpop/data_processing/encoders.py
@@ -15,9 +15,7 @@ import numpy.typing as npt
 from synthpop.utils import str_dtype, to_missing_str_array
 
 
-def validate_string_array(x):
-    arr = to_missing_str_array(x)
-
+def to_1D(arr):
     if arr.ndim >1:
         if arr.shape[1] !=1:
             raise ValueError("input should be 1D")
@@ -25,6 +23,11 @@ def validate_string_array(x):
             return arr.reshape(-1)
             
     return arr
+    
+def validate_string_array(x):
+    arr = to_missing_str_array(x)
+            
+    return to_1D(arr)
 
 class PCAEncoder(TransformerMixin, BaseEstimator):
     """
@@ -122,11 +125,6 @@ class PCAEncoder(TransformerMixin, BaseEstimator):
         X_val = validate_string_array(X)#to_missing_str_array(X).reshape(X.shape[0])
         y_val = validate_string_array(y)#to_missing_str_array(y).reshape(y.shape[0])
 
-
-        # if X_val.ndim != 1:
-        #     raise ValueError("X should be 1D")
-        # if y_val.ndim != 1:
-        #     raise ValueError("Y should be 1D")
         
         if X_val.shape[0] == 0 and y_val.shape[0]==0:
             self.mapping_ = {}
@@ -192,17 +190,18 @@ class PCAEncoder(TransformerMixin, BaseEstimator):
         #     return np.zeros(X.shape[0])
         missing_mapping =  { np.nan:[np.nan]*self.n_features_out_}
         mapping_including_missing = self.mapping_|missing_mapping if hasattr(self,"mapping_") else missing_mapping
+        X_val = validate_string_array(X)
 
         # if X contains Nones, then the dtype of X is object.
         # In that case, X cannot be sorted.
         # many routines of numpy for finding differences between lists depend on the items being sortable.
         
         if hasattr(self,"mapping_"):
-            unique_values_in_x =set(X[~pd.isna(X)])#np.unique([v for v in X if not pd.isna(v)])
+            unique_values_in_x =set(X_val[~pd.isna(X_val)])#np.unique([v for v in X if not pd.isna(v)])
             if not unique_values_in_x.issubset(self.mapping_.keys()):
                 raise ValueError("new values not seen during fitting when encoding.")
 
-        return np.array([self.mapping_[v] if v==v else [np.nan]* self.n_features_out_ for v in X],dtype=np.float32)
+        return np.array([self.mapping_[v] if v==v else [np.nan]* self.n_features_out_ for v in X_val],dtype=np.float32)
 
    
 class MeanEncoder(OneToOneFeatureMixin,TransformerMixin, BaseEstimator): 
@@ -258,8 +257,8 @@ class MeanEncoder(OneToOneFeatureMixin,TransformerMixin, BaseEstimator):
 
         if X.shape[0] == 0 or y.shape[0] == 0:
             raise ValueError("mean encoding not possible for empty arrays")
-        X_val =  to_missing_str_array(X)
-        y_val = y
+        X_val =  validate_string_array(X)
+        y_val = to_1D(y)
 
         self.n_features_in_ = 1
 
@@ -302,27 +301,25 @@ class MeanEncoder(OneToOneFeatureMixin,TransformerMixin, BaseEstimator):
         """
 
         check_is_fitted(self, 'mapping_')
+        X_val = validate_string_array(X)
 
-        # Input validation
-        #X = np.asarray(X)
+        result = np.full((len(X_val),1), np.nan, dtype=np.float32)
 
-        if X.ndim != 1:
-            raise ValueError(f"X must be a 1D array, got shape {X.shape}.")
+
         if len(self.mapping_) == 0:
-            return np.full(len(X), np.nan, dtype=np.float32).reshape(-1, 1) #2D output with only nans
+            return result #2D output with only nans
         
         # Start transform
-        result = np.full(len(X), np.nan, dtype=np.float32)
 
-        X_missing = np.isnan(X)
+        X_missing = np.isnan(X_val)
         
         # Detect unseen categories
-        unseen_categories = set(X[~X_missing]) - set(self.mapping_.keys())
+        unseen_categories = set(X_val[~X_missing]) - set(self.mapping_.keys())
         if unseen_categories:
             raise ValueError(f"Column to be encoded X has unseen categories: {unseen_categories}")
         
         # Apply mapping
-        for i, val in enumerate(X):
+        for i, val in enumerate(X_val):
             if not X_missing[i]:
                 result[i] = self.mapping_[val]
 
--- a/src/synthpop/data_processing/encoders.py
+++ b/src/synthpop/data_processing/encoders.py
@@ -31,7 +31,7 @@ class _BaseEncoder(TransformerMixin, BaseEstimator):
         """
         if arr.ndim >1:
             if arr.shape[1] !=1:
-                raise ValueError("input should be 1D or be 2D with one column")
+                raise ValueError(f"Expected a 1D or a 2D array with exactly one column. Received shape {arr.shape}.")
             else:
                 return arr.reshape(-1)
             
@@ -51,7 +51,9 @@ class _BaseEncoder(TransformerMixin, BaseEstimator):
        X_missing = np.isnan(X_val)
         
        # Detect unseen categories
-       unseen_categories = np.setdiff1d(X_val[~X_missing], list(self.mapping_.keys()))
+       seen = set(self.mapping_.keys())
+observed = set(X_val[~X_missing])
+unseen = observed - seen
        if unseen_categories.size >0:
            raise ValueError(f"transform received unseen categories. Unseen values: {unseen_categories}. Ensure input was fitted")
 
@@ -60,7 +62,7 @@ class _BaseEncoder(TransformerMixin, BaseEstimator):
         if X_val.shape[0] == 0:
             return np.empty(shape= (0,self.n_features_out_),dtype=np.float32)
  
-        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32)#preallocation
+        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32) #pre-allocation
 
         X_missing = np.isnan(X_val)
 
--- a/src/synthpop/data_processing/encoders.py
+++ b/src/synthpop/data_processing/encoders.py
@@ -160,7 +160,7 @@ class PCAEncoder(_BaseEncoder):
         tags.estimator_type = "transformer"
         return tags
 
-    def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
+    def fit(self,X:npt.NDArray, y: npt.NDArray) -> Self:
         """
         Calculate the encoding.
         
@@ -227,7 +227,7 @@ class PCAEncoder(_BaseEncoder):
 
         return self
 
-    def transform(self, X: npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optimal for decision trees.
+    def transform(self, X: npt.NDArray) -> npt.NDArray[np.float32]:#float32 is optimal for decision trees.
         """
         replaces each level of `X` with the numerical values determined in :py:meth:`fit`
 
@@ -272,7 +272,7 @@ class MeanEncoder(_BaseEncoder):
         tags.estimator_type = "transformer"
         return tags
 
-    def fit(self, X: npt.ArrayLike, y: npt.ArrayLike) -> Self:
+    def fit(self, X: npt.NDArray, y: npt.NDArray) -> Self:
         """
         Calculate average y value for each X category.
         
@@ -323,7 +323,7 @@ class MeanEncoder(_BaseEncoder):
 
         return self
 
-    def transform(self, X: npt.ArrayLike) -> npt.NDArray[np.float32]:#float32 is optimal for decision trees.
+    def transform(self, X: npt.NDArray) -> npt.NDArray[np.float32]:#float32 is optimal for decision trees.
         """
         Apply mapping from fitting function to ``X`` and returns the encoded version ``X_transformed``
         
--- a/src/synthpop/data_processing/encoders.py
+++ b/src/synthpop/data_processing/encoders.py
@@ -12,6 +12,7 @@ import pandas as pd
 import numpy as np
 import numpy.typing as npt
 
+from synthpop.utils import str_dtype, to_missing_str_array
 
 
 class PCAEncoder(TransformerMixin, BaseEstimator):
@@ -263,29 +264,26 @@ class MeanEncoder(OneToOneFeatureMixin,TransformerMixin, BaseEstimator):
             >>> encoder.fit(X, y)
         """
 
-        # Required sklearn attributes
-        y = np.array([np.nan if (v is pd.NA ) else v for v in y], dtype=float) #for pd.NA compatibility
+        if not np.issubdtype(y.dtype,np.number):
+            raise ValueError("target must be numeric dtype for mean encoding")
 
-        X_val, y_val = validate_data(self, X=X, y=y, validate_separately = (
-             dict(ensure_2d=False, ensure_min_samples=1, dtype=["str", "object"], ensure_all_finite="allow-nan"),
-             dict(ensure_2d=False, ensure_min_samples=1, dtype='numeric', ensure_all_finite="allow-nan")
-        ))
+        if X.shape[0] == 0 or y.shape[0] == 0:
+            raise ValueError("mean encoding not possible for empty arrays")
+        X_val =  to_missing_str_array(X)
+        y_val = y
 
-        if isinstance(X,pd.Series) and X.name is not None:
-            self.feature_names_in_ = [X.name]
         self.n_features_in_ = 1
 
 
         # Identify missing
-        X_missing = np.zeros(len(X_val), dtype=bool)
-        X_missing |= pd.isna(X_val)
-        y_missing = pd.isna(y_val)
+        X_missing = np.isnan(X_val)
+        y_missing = np.isnan(y_val)
         
         # Fit encoder
         self.mapping_ = {}
-        unique_categories = np.unique(X_val[~X_missing].astype(str))
+        unique_categories = np.unique(X_val[~X_missing])
         for cat in unique_categories:
-            mask = (~X_missing) & (X_val.astype(str) == cat)
+            mask = (~X_missing) & (X_val == cat)
             valid_targets = y_val[mask & ~y_missing]
             if valid_targets.size == 0:
                 mean_val = np.nan
@@ -317,7 +315,7 @@ class MeanEncoder(OneToOneFeatureMixin,TransformerMixin, BaseEstimator):
         check_is_fitted(self, 'mapping_')
 
         # Input validation
-        X = np.asarray(X)
+        #X = np.asarray(X)
 
         if X.ndim != 1:
             raise ValueError(f"X must be a 1D array, got shape {X.shape}.")
@@ -327,11 +325,7 @@ class MeanEncoder(OneToOneFeatureMixin,TransformerMixin, BaseEstimator):
         # Start transform
         result = np.full(len(X), np.nan, dtype=np.float32)
 
-        # Detect missing
-        X_missing = np.zeros(len(X), dtype=bool)
-        if X.dtype.kind in ("f", "i"):
-            X_missing |= np.isnan(X)
-        X_missing |= np.equal(X, None)
+        X_missing = np.isnan(X)
         
         # Detect unseen categories
         unseen_categories = set(X[~X_missing]) - set(self.mapping_.keys())
@@ -344,18 +338,3 @@ class MeanEncoder(OneToOneFeatureMixin,TransformerMixin, BaseEstimator):
                 result[i] = self.mapping_[val]
 
         return result
-    
-    def get_feature_names_out(self, input_features=None):
-        check_is_fitted(self, "mapping_")
-
-        if input_features is None:
-            if hasattr(self, "feature_names_in_"):
-                input_features = self.feature_names_in_
-            else:
-                input_features = [f"mean_x{i}" for i in range(self.n_features_in_)]
-
-        if hasattr(self, "feature_names_in_"):
-            if not np.array_equal(input_features, self.feature_names_in_):
-                raise ValueError(f"input_features must match feature_names_in_. Expected {self.feature_names_in_}, got {input_features}")
-        base = input_features[0]
-        return np.asarray([f"{base}_mean"], dtype=object)
--- a/src/synthpop/data_processing/encoders.py
+++ b/src/synthpop/data_processing/encoders.py
@@ -40,7 +40,7 @@ class _BaseEncoder(TransformerMixin, BaseEstimator):
     def validate_string_array(self,x):
         
         """
-        Transform all missing values in a string array to np.nan.
+        Transform all missing values in a string array to np.nan. It also converts dtype, reshapes and validates dimensionality.
         :param x: an array of strings
         :return: the 1-dimensional array of strings with one value for missing
         """
@@ -52,8 +52,8 @@ class _BaseEncoder(TransformerMixin, BaseEstimator):
         
        # Detect unseen categories
        seen = set(self.mapping_.keys())
-observed = set(X_val[~X_missing])
-unseen = observed - seen
+       observed = set(X_val[~X_missing])
+       unseen = observed - seen
        if unseen:
            raise ValueError(f"transform received categories that were not observed during fitting. Unseen values: {sorted(unseen)}. Ensure input was fitted")
 
@@ -82,7 +82,7 @@ unseen = observed - seen
 
         result = mapped_vals[rev_index]
 
-        result[X_missing] = [np.nan]*self.n_features_out_
+        result[X_missing,:] = np.nan#[np.nan]*self.n_features_out_
 
         if result.ndim == 1:
             return result.reshape((-1,1))
@@ -217,7 +217,7 @@ class PCAEncoder(_BaseEncoder):
         if isinstance(pca_result,pd.DataFrame):
             pca_result = pca_result.to_numpy()
 
-        self.n_features_out_ = np.atleast_2d(pca_result) #needed for transform
+        self.n_features_out_ = pca_result.shape[1] #needed for transform
     
         mapping_for_missing = {k:[np.nan]*self.n_features_out_ for k in x_such_that_y_is_always_missing}#The values of X s.t. y is always missing
 
--- a/src/synthpop/data_processing/encoders.py
+++ b/src/synthpop/data_processing/encoders.py
@@ -108,13 +108,10 @@ class PCAEncoder(TransformerMixin, BaseEstimator):
 
         self.n_features_in_ = 1
 
-        X_val,y_val = validate_data(self,X=X,y=y, validate_separately = (
-            dict(ensure_2d=False,dtype=["str","object"],ensure_all_finite="allow-nan",ensure_min_samples=0)
-            ,dict(ensure_2d=False,dtype=["str","object"],ensure_all_finite="allow-nan",ensure_min_samples=0)
-            ))
+        X_val = to_missing_str_array(X)
+        y_val = to_missing_str_array(y)
 
-        if isinstance(X,pd.Series):
-            self.feature_names_in_ = [X.name]
+
         if X_val.ndim != 1:
             raise ValueError("X should be 1D")
         if y_val.ndim != 1:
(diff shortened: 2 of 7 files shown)


src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,58 @@
...
+    def to_1D(self,arr):
+        if arr.ndim >1:
+            if arr.shape[1] !=1:
+                raise ValueError("input should be 1D")
+            else:
+                return arr.reshape(-1)
+            
+        return arr
+
+    def validate_string_array(self,x):
```
anneliesdej, line 40:
```suggestion
    def validate_string_array(self,x):
    """
    Transform all missing values in a string array to np.nan.
    :param x: an array of strings
    :return: the 1-dimensional array of strings with one value for missing
    """
```
Even though it is internal, it would be better to explain slightly what each function does/what the goal is/why it is needed

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,58 @@
...
 import numpy.typing as npt
 
+from synthpop.utils import str_dtype, to_missing_str_array
 
 
-class PCAEncoder(TransformerMixin, BaseEstimator):
+
+
+class _BaseEncoder(TransformerMixin, BaseEstimator):
+    def to_1D(self,arr):
```
anneliesdej, line 21: Here some comment is also needed as it does not immediately feel intuitive that it is an error or a change to the data

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,58 @@
...
+
+    def _check_unseen_values(self,X_val):
+       X_missing = np.isnan(X_val)
+        
+       # Detect unseen categories
+       unseen_categories = set(X_val[~X_missing]) - set(self.mapping_.keys())
+       if unseen_categories:
+           raise ValueError(f"Column to be encoded X has unseen categories: {unseen_categories}")
+
+    def _apply_mapping(self,X_val):
```
anneliesdej, line 60: There are too many white lines in this function ;)

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,58 @@
...
+
+        
+        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32)
+
+        if not hasattr(self,"mapping_"):
+            return np.full((len(X_val),1), np.nan, dtype=np.float32)
+
+
+        if len(self.mapping_) == 0:
+            return result #2D output with only nans
```
anneliesdej: Documentation, in probably the doc strings, is needed to explain why we want 1D input and the output is 2D

src/synthpop/utils.py
```diff
@@ -0,0 +1,8 @@
+import numpy as np
+
+str_dtype = np.dtypes.StringDType(na_object=np.nan)
+
+def to_missing_str_array(x):
```
anneliesdej:
```suggestion
def to_stringdtype_array(x):
```
The current name is not very clear on what happens in the function

anneliesdej (commented):
Review on the draft: current simplifying seems to head in the right direction

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,70 @@
...
+        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32)#preallocation
+
+        X_missing = np.isnan(X_val)
+
+        unique_vals = np.unique(X_val,return_inverse=False)#The reverse_index does not work well with nan. 
+        rev_index = np.searchsorted(unique_vals, X_val)
+
+        
+        #checking for np.nan in a list comprehension is complicated
+        # np.isnan does not work well for stringDType.
```
anneliesdej: if np.isnan does not work well, would it be better to fall back to pd.isna?

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,70 @@
...
+        # v==np.nan will always be false, even if v is np.nan.
+        # Since v==v is only false if v = nan, this seems like the simplest and most reliable way to solve this.
+        # using nan as a key in the dictionary does not work either, probably for the same reasons.
+        mapped_vals = np.array([self.mapping_[v] if v==v else [np.nan]*self.n_features_out_ for v in unique_vals],dtype=np.float32)
+
+        result = mapped_vals[rev_index]
```
anneliesdej, line 83: Right now category-level processing is mixed with row-leven indexing in a way that can lead to incorrect handling of missing. `mapped_vals` is made with `unique_vals`, but indexed using `rev_index` that was computed over the full `X_val` which includes missing entries which can mismatch.

Additionally, missing values are only considered during mapping construction but are not explicitly enforced in the final output. This means NaN rows can accidentally inherit valid encoded values depending on how indexing resolves.

A more robust approach would be to explicitly set `reesult[X_missing] = np.nan` after mapping. This separates category mapping from row-level missing value handling

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,70 @@
...
+           raise ValueError(f"Column to be encoded X has unseen categories: {unseen_categories}")
+
+    def _apply_mapping(self,X_val):
+        
+        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32)#preallocation
+
+        X_missing = np.isnan(X_val)
+
+        unique_vals = np.unique(X_val,return_inverse=False)#The reverse_index does not work well with nan. 
+        rev_index = np.searchsorted(unique_vals, X_val)
```
anneliesdej: `searchsorted` assumes sorted order, but nans break this ordering guarantee. Better might be something like 
```python
unique_vals, rev_index = np.unique(X_val, return_inverse=True)
```

src/synthpop/data_processing/encoders.py
```diff
@@ -107,30 +156,20 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
+            raise ValueError("pca encoding not possible for empty arrays")
 
         # the core of this implementation
 
 
         #The alternative to using pandas here is either use scipy or DIY.
         missing_contingency_table = pd.crosstab(X_val,pd.isna(y_val))
-        if not False in missing_contingency_table:
+
+        if not False in missing_contingency_table:# If there are only missing values for y:
```
anneliesdej:
```suggestion
        if True in missing_contingency_table.columns:
```

Or `if (~missing_contingency_table.columns).any():`

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,70 @@
...
+        unique_vals = np.unique(X_val,return_inverse=False)#The reverse_index does not work well with nan. 
+        rev_index = np.searchsorted(unique_vals, X_val)
+
+        
+        #checking for np.nan in a list comprehension is complicated
+        # np.isnan does not work well for stringDType.
+        # v==np.nan will always be false, even if v is np.nan.
+        # Since v==v is only false if v = nan, this seems like the simplest and most reliable way to solve this.
+        # using nan as a key in the dictionary does not work either, probably for the same reasons.
+        mapped_vals = np.array([self.mapping_[v] if v==v else [np.nan]*self.n_features_out_ for v in unique_vals],dtype=np.float32)
```
anneliesdej: not sure if we should use a "hack" like this

src/synthpop/data_processing/encoders.py
```diff
@@ -107,30 +156,20 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
-        if X_val.shape[0] == 0 and y_val.shape[0]==0:
-            self.mapping_ = {}
-            self.n_features_out_ = 0
+        X_val = self.validate_string_array(X)
+        y_val = self.validate_string_array(y)
 
-            return self
+        
+        if X_val.shape[0] == 0 or y_val.shape[0]==0:
+            raise ValueError("pca encoding not possible for empty arrays")
```
anneliesdej:
```suggestion
            raise ValueError("Cannot fit encoder: X and y must be non-empty.")
```
This gives a clearer API-level contract

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,70 @@
...
+       # Detect unseen categories
+       unseen_categories = np.setdiff1d(X_val[~X_missing], list(self.mapping_.keys()))
+       if unseen_categories.size >0:
+           raise ValueError(f"Column to be encoded X has unseen categories: {unseen_categories}")
+
+    def _apply_mapping(self,X_val):
+        
+        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32)#preallocation
+
+        X_missing = np.isnan(X_val)
```
anneliesdej, line 67: X_missing is not used right now

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,70 @@
...
+        arr = to_stringdtype_array(x)
+        return self.to_1D(arr)
+
+    def _check_unseen_values(self,X_val):
+       X_missing = np.isnan(X_val)
+        
+       # Detect unseen categories
+       unseen_categories = np.setdiff1d(X_val[~X_missing], list(self.mapping_.keys()))
+       if unseen_categories.size >0:
+           raise ValueError(f"Column to be encoded X has unseen categories: {unseen_categories}")
```
anneliesdej:
```suggestion
           raise ValueError(f"PCAEncoder.transform received unseen categories. Unseen values: {unseen_categories}. Ensure input was fitted")
```
This makes the error message more actionable

src/synthpop/data_processing/encoders.py
```diff
@@ -263,36 +271,34 @@ def fit(self, X: npt.ArrayLike, y: npt.ArrayLike) -> Self:
             >>> encoder.fit(X, y)
         """
 
-        # Required sklearn attributes
-        y = np.array([np.nan if (v is pd.NA ) else v for v in y], dtype=float) #for pd.NA compatibility
+        if not np.issubdtype(y.dtype,np.number):
+            raise ValueError("target must be numeric dtype for mean encoding")
```
anneliesdej:
```suggestion
            raise ValueError(f"MeanEncoder requires numeric target array y. Received dtype={y.dtype}")
```

src/synthpop/data_processing/encoders.py
```diff
@@ -263,36 +271,34 @@ def fit(self, X: npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
 
-        X_val, y_val = validate_data(self, X=X, y=y, validate_separately = (
-             dict(ensure_2d=False, ensure_min_samples=1, dtype=["str", "object"], ensure_all_finite="allow-nan"),
-             dict(ensure_2d=False, ensure_min_samples=1, dtype='numeric', ensure_all_finite="allow-nan")
-        ))
+        if X.shape[0] == 0 or y.shape[0] == 0:
+            raise ValueError("mean encoding not possible for empty arrays")
```
anneliesdej:
```suggestion
            raise ValueError("Cannot fit encoder: X and y must be non-empty.")
```

src/synthpop/data_processing/encoders.py
```diff
@@ -263,36 +271,34 @@ def fit(self, X: npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
-            mask = (~X_missing) & (X_val.astype(str) == cat)
+            mask = (~X_missing) & (X_val == cat)
             valid_targets = y_val[mask & ~y_missing]
             if valid_targets.size == 0:
                 mean_val = np.nan
             else:
                 mean_val = valid_targets.mean()
             
-            self.mapping_[cat] = np.float32(mean_val)
+            self.mapping_[cat] = [np.float32(mean_val)]
```
anneliesdej: we could also do np.mean(valid_targets, dtype=np.float32). This prevents the casting to float64 first

anneliesdej (changes requested):

src/synthpop/data_processing/encoders.py
KoomenErina: `np.unique` act suprising to me when dealing with an StringDType array:

```python
import numpy as np
str_dtype = np.dtypes.StringDType(na_object=np.nan)
x_val = np.array(["a", "a", "b", "b", "c",np.nan, "c"],dtype = str_dtype)
unique_vals,rev_index = np.unique(x_val,return_inverse=True)
unique_vals
```
>array(['a', 'b', 'c'], dtype=StringDType(na_object=nan))
```python
rev_index
```
>array([0, 0, 1, 1, 2, 2, 2])

```python
unique_vals[rev_index]
```
>array(['a', 'a', 'b', 'b', 'c', 'c', 'c'],
      dtype=StringDType(na_object=nan))

The [docs ](https://numpy.org/doc/2.4/reference/generated/numpy.unique.html) of np.unique explicitly mention that rev_index could be used to reconstruct the original input. That does not seem to hold here.

KoomenErina: for now, the code deals with this property

src/synthpop/data_processing/encoders.py
KoomenErina: The condition in the `if` statement is intended to say: "if every value of `y` is missing". This is formulated as "there is no row in `y` that is not missing". The suggestions are not equivalent to this.

I have adapted the last suggestion

src/synthpop/data_processing/encoders.py
```diff
@@ -263,36 +278,30 @@ def fit(self, X: npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
-            mask = (~X_missing) & (X_val.astype(str) == cat)
+            mask = (~X_missing) & (X_val == cat)
             valid_targets = y_val[mask & ~y_missing]
-            if valid_targets.size == 0:
-                mean_val = np.nan
-            else:
-                mean_val = valid_targets.mean()
             
-            self.mapping_[cat] = np.float32(mean_val)
+            self.mapping_[cat] = np.mean(valid_targets, dtype=np.float32)#[np.float32(mean_val)]
```
anneliesdej:
```suggestion
            self.mapping_[cat] = np.array([np.mean(valid_targets, dtype=np.float32)]).astype(np.float32)
```
For coherence, this would return a vector instead of a scalar. In PCA it is also a vector

src/synthpop/data_processing/encoders.py
```diff
@@ -263,36 +278,30 @@ def fit(self, X: npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
-        unique_categories = np.unique(X_val[~X_missing].astype(str))
+        unique_categories = np.unique(X_val[~X_missing])
         for cat in unique_categories:
-            mask = (~X_missing) & (X_val.astype(str) == cat)
+            mask = (~X_missing) & (X_val == cat)
             valid_targets = y_val[mask & ~y_missing]
-            if valid_targets.size == 0:
-                mean_val = np.nan
-            else:
-                mean_val = valid_targets.mean()
```
anneliesdej: I would restore the explicit handling of if valid_targets.size == 0, because right now you get a RunTimeWarning if that is the case. Functionally it is OK, but it is preferable to not have external warnings when running the package

src/synthpop/data_processing/encoders.py
anneliesdej:
```suggestion
from sklearn.base import TransformerMixin, BaseEstimator
```
OneToOneFeatureMixin is unused

anneliesdej:
```suggestion
from sklearn.utils.validation import check_is_fitted
```
validate_data is unused

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,77 @@
 import numpy as np
 import numpy.typing as npt
 
+from synthpop.utils import str_dtype,to_stringdtype_array
```
anneliesdej:
```suggestion
from synthpop.utils import to_stringdtype_array
```
str_dtype is unused

src/synthpop/data_processing/encoders.py
```diff
@@ -107,30 +163,20 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
-        
-        if X_val.shape[0] == 0 and y_val.shape[0]==0:
-            self.mapping_ = {}
-            self.n_features_out_ = 0
+        X_val = self.validate_string_array(X)
+        y_val = self.validate_string_array(y)
 
-            return self
+        
+        if X_val.shape[0] == 0 or y_val.shape[0]==0:
```
anneliesdej, line 178: There could also be a check if X_val.shap[0] != y_val.shape[0] which raises a ValueError to prevent pandas crosstab to break. The break then happens explicitly in the package.

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,77 @@
...
+        if result.ndim == 1:
+            return result.reshape((-1,1))
+
+        return result
+        
+
+class PCAEncoder(_BaseEncoder):
     """
     Transforms categorical data to one or more numeric columns.
     The user can adjust the amount of principle components by passing an instance of sklearn.decomposition.PCA to `pca_transform`
```
anneliesdej:
```suggestion
    The user can adjust the amount of principal components by passing an instance of sklearn.decomposition.PCA to `pca_transform`
```

src/synthpop/data_processing/encoders.py
```diff
@@ -39,18 +107,6 @@ class PCAEncoder(TransformerMixin, BaseEstimator):
...
-        >>> encoder = PCAEncoder().set_output(transform="pandas")
-        >>> encoder.fit_transform(X=X,y=y)
-        input_feature_pca0  input_feature_pca1  input_feature_pca2
-        0           -1.118034       -1.500000e+00       -1.201987e-16
-        1           -1.118034       -1.500000e+00       -1.201987e-16
-        2            2.236068       -1.295326e-15       -1.201987e-16
-        3            2.236068       -1.295326e-15       -1.201987e-16
-        4           -1.118034        1.500000e+00       -1.201987e-16
     
         with a different number of principle components (only the first):
```
anneliesdej:
```suggestion
        with a different number of principal components (only the first):
```

src/synthpop/data_processing/encoders.py
anneliesdej:
```suggestion
         See `sklearn.decomposition.PCA <https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html>`_ for the possible parameters. With the default parameters, all principal components are computed and used.
```

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,77 @@
...
 
 
-class PCAEncoder(TransformerMixin, BaseEstimator):
+
+
+class _BaseEncoder(TransformerMixin, BaseEstimator):
+    def to_1D(self,arr):
+        """
+        The expected input of the encoders is 2D.
+        Conceptually, the encoders work on a per-column basis.
```
anneliesdej:
```suggestion
        Conceptually, the encoders operate on a single feature column at a time.
```

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,77 @@
...
+        That is why we need to convert to 1D arrays.
```
anneliesdej:
```suggestion
        This helper converts input to a 1-dimensional array while allowing both:
        - 1D arrays of shape (n_samples,)
        - 2D single-column arrays of shape (n_samples, 1)
        
        Arrays with more than one column are rejected.
                
```

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,77 @@
...
+       unseen_categories = np.setdiff1d(X_val[~X_missing], list(self.mapping_.keys()))
+       if unseen_categories.size >0:
+           raise ValueError(f"transform received unseen categories. Unseen values: {unseen_categories}. Ensure input was fitted")
+
+    def _apply_mapping(self,X_val):
+
+        if X_val.shape[0] == 0:
+            return np.empty(shape= (0,self.n_features_out_),dtype=np.float32)
+
+
```
anneliesdej:
```suggestion
```
Remove white line

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,77 @@
...
+        
+        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32)#preallocation
+
+        X_missing = np.isnan(X_val)
+
+        if X_missing.all():
+            return result
+
+        unique_vals,rev_index = np.unique(X_val,return_inverse=True)#The reverse_index does not work well with nan.
```
anneliesdej: Is the comment about reverse_index still needed?

src/synthpop/data_processing/encoders.py
```diff
@@ -12,9 +12,77 @@
...
+        #Note that rev_index does not reconstruct X_val when X_val is a stringDType array with np.nan:
```
anneliesdej, line 73: Are these comments still needed?

tests/unit/test_mean_encoder.py
```diff
@@ -73,55 +72,49 @@ def test_fit_multiple_times():
...
     enc.fit(X2, y2)
     
-    assert enc.mapping_["a"] == 10, "Refitting does not work properly"
-    assert enc.mapping_["b"] == 20, "Refitting does not work properly"
+    assert enc.mapping_["a"] == [10], "Refitting does not work properly"
+    assert enc.mapping_["b"] == [20], "Refitting does not work properly"
 
 def test_fit_with_empty_arrays():
+    """
+    For now, we don't have a clear usage scenario in which it would make sense to define what happens when the input is an empty array.
```
anneliesdej:
```suggestion
    For now, we do not have a clear usage scenario in which empty inputs should be supported.
```

tests/unit/test_mean_encoder.py
```diff
@@ -73,55 +72,49 @@ def test_fit_multiple_times():
...
+    Altough empty array => empty mapping_ would make sense, it is more likely to hide a serious bug than to help the user for now.
```
anneliesdej:
```suggestion
    Although outputting an empty `mapping_` for an empty array makes sense, it is more likely to hide a serious bug than to help the user for now.
    This tet ensures that poviding empty arrays raises a ValueError, since the expected behaviour for empty datasets is currently undefined in the functional description.
```

tests/unit/test_pca_encoder.py
```diff
@@ -58,52 +59,27 @@ def get_test_data_full():
...
 
 
 def get_test_data_feature_constants():
     #   constant feature
-    X = np.array(["a", "a", "a", "a"])
-    y = np.array(["x", "x", "y", "z"])
+    X = np.array(["a", "a", "a", "a"],dtype = str_dtype)
+    y = np.array(["x", "x", "y", "z"],dtype = str_dtype)
 
     expected_input_pca = np.array([  # centred table       contingency table
```
anneliesdej: Is this added to show an assumption? Asking since it is not used.

tests/unit/test_pca_encoder.py
```diff
@@ -230,47 +201,17 @@ def test_pca_fit_constant_feature():
...
-    assert_dict(expected_dict, result.mapping_)
-    validate_set_inout_count(result, expected_n_feat=len(expected_dict["a"]))
-
-    assert result.feature_names_in_ == ["input_feature"]
 
 
 def test_pca_fit_empty_input():
     """
-    fitting on empty data should result in an empty mapping.
+    For now, we don't have a clear usage scenario in which it would make sense to define what happens when the input is an empty array.
```
anneliesdej: Update this also to show the same as the text in the MeanEncoder tests

tests/unit/test_pca_encoder.py
```diff
@@ -375,27 +288,19 @@ def test_pca_transform_empty_input():
...
-
-    result = encoder.transform(X=np.array([]))
-    assert len(result) == 0
-
-@pytest.mark.parametrize("missing", [None,pd.NA,np.nan])
+@pytest.mark.parametrize("missing", [np.nan])
 def test_pca_transform_when_mapping_is_empty_transform_missing_to_nan(missing):
     """
-    transforming an empty X should always result in an empty array, even if the mapping is empty.
+    transforming an always missing X should always result in an always nan array, even if the mapping is empty.
```
anneliesdej:
```suggestion
    transforming an always missing X should result in an always nan array, even if the mapping is empty.
```

tests/unit/test_pca_encoder.py
```diff
@@ -410,15 +315,15 @@ def test_pca_transform_exception_on_new_value():
     encoder.mapping_ = {"a": [1.1, 2.2], "c":[3.3,4.4]}
     encoder.n_features_out_ = 2
 
-    with pytest.raises(ValueError, match="values not seen during fitting"):
+    with pytest.raises(ValueError, match="transform received unseen categories. Unseen values:"):
         result = encoder.transform(np.array(["b","c"]))
```
anneliesdej:
```suggestion
        encoder.transform(np.array(["b","c"]))
```

tests/integration/test_int_pca_encoding.py
```diff
@@ -74,32 +76,29 @@ def test_pca_encoding_fit_constant_feature():
     # The behaviour is different as described in the functional descriptions.
```
anneliesdej, line 73: Is this still the case? If yes, code or functional description needs to be updated

tests/integration/test_int_pca_encoding.py
```diff
@@ -74,32 +76,29 @@ def test_pca_encoding_fit_constant_feature():
...
-    encoder = PCAEncoder()
-    result = encoder.fit_transform(X,y)
-    assert not np.equal(result[0],result[1]).all()
+    before = encoder.fit_transform(X,y)
+    
+    #Using the setoutput api here via a context manager could have effect on sklearn.PCA
+    with config_context(transform_output="pandas"):
+        after = encoder.fit_transform(X,y)
+
+    assert np.array_equal(before,after), "set output api break pca encoding"
```
anneliesdej:
```suggestion
    assert np.array_equal(before,after), "Scikit-learn's set_output API breaks PCA encoding"
```

anneliesdej (changes requested):

src/synthpop/data_processing/encoders.py
KoomenErina: I do think so. A future developer might expect that `result` on line 75 contains the correct values, since that is basically what the numpy docs imply. If that developer would operate under that assumption they would not get an error but silently the wrong result. 

This would result in the kind of bugs that are really hard to diagnose.

Is it clear that these comments are to inform future developers? or do they look like "temporarly removed" code?

tests/unit/test_pca_encoder.py
KoomenErina: It is added to test that this function works with `StringDType`

src/synthpop/data_processing/encoders.py
KoomenErina: I see that pandas raises the exception "ValueError: All arrays must be of the same length".  I see how this can be confusing for users of this package. They might be confused about which arrays need to be the same length.

src/synthpop/data_processing/encoders.py
anneliesdej: Adding "Note" makes it look more like information than temporarily removed code. However, I would add still why rev_index is used then if it does not seem to work correctly

tests/unit/test_pca_encoder.py
KoomenErina: the function is not used anymore, so it has been removed

docs/source/functional descriptions/PCA-encoding.md
```diff
@@ -51,7 +51,7 @@ Only the first $k$ principal components are kept as part of the encoding.
 |missing | missing | missing|
 |any specific non-missing value| always missing for that specific feature value | missing|
 |any specific non-missing value| sometimes but not always missing for that specific feature value|the encoding treats the missing target value as a normal value and the default PCA is applied
-|many different non-missing values| the same constant over all values of the feature (any non-missing value)| The rotation matrix of PCA becomes an identity matrix. The contingency table becomes a 1D vector of the number of occurrences of the feature level. |
+|many different non-missing values| the same constant over all values of the feature (any non-missing value)| The rotation matrix of PCA becomes an identity matrix. The contingency table becomes a 1D vector of the number of occurrences of the feature level. The encoding is an scaled and centered version of count encoding |
```
anneliesdej:
```suggestion
|many different non-missing values| the same constant over all values of the feature (any non-missing value)| The rotation matrix of PCA becomes an identity matrix. The contingency table becomes a 1D vector of the number of occurrences of the feature level. The encoding is a scaled and centred version of count encoding |
```
Corrected for British English

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,95 @@
...
+        # x_val = np.array(["a", "a", "b", "b", "c",np.nan, "c"],dtype = str_dtype)
+        # unique_vals,rev_index = np.unique(x_val,return_inverse=True)
+        # unique_vals[rev_index]
+        # array(['a', 'a', 'b', 'b', 'c', 'c', 'c'],dtype=StringDType(na_object=nan))
+        
+        mapped_vals = np.array([self.mapping_[v] for v in unique_vals],dtype=np.float32)
+
+        result = mapped_vals[rev_index]
+
+        result[X_missing] = [np.nan]*self.n_features_out_
```
anneliesdej:
```suggestion
        result[X_missing] = np.nan
```
Seems safer and more natural language wise

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,95 @@
...
+        :return: the 1-dimensional array of strings with one value for missing
+        """
+        arr = to_stringdtype_array(x)
+        return self.to_1D(arr)
+
+    def _check_unseen_values(self,X_val):
+       X_missing = np.isnan(X_val)
+        
+       # Detect unseen categories
+       unseen_categories = np.setdiff1d(X_val[~X_missing], list(self.mapping_.keys()))
```
anneliesdej:
```suggestion
       seen = set(self.mapping_.keys())
observed = set(X_val[~X_missing])
unseen = observed - seen
```
np.setdiff1d silently casts types and can make nan handling inconsistent. While it is not a big issue for filtering, to prevent a false error this code might be better

src/synthpop/data_processing/encoders.py
```diff
@@ -263,36 +284,38 @@ def fit(self, X: npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
+            mask = (~X_missing) & (X_val == cat)
             valid_targets = y_val[mask & ~y_missing]
             if valid_targets.size == 0:
-                mean_val = np.nan
+                self.mapping_[cat] = np.array([np.nan],dtype=np.float32)
             else:
-                mean_val = valid_targets.mean()
-            
-            self.mapping_[cat] = np.float32(mean_val)
+                self.mapping_[cat] = np.mean(valid_targets, dtype=np.float32,keepdims=True)
```
anneliesdej, line 322:
```suggestion
                self.mapping_[cat] = np.array([np.mean(valid_targets], dtype=np.float32
```
Apparently np.mean does not have as expected for all NumPy versions when it comes to dtype. Not sure if we need to faciliate it tho

src/synthpop/data_processing/encoders.py
```diff
@@ -161,7 +213,7 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
         if isinstance(pca_result,pd.DataFrame):
             pca_result = pca_result.to_numpy()
 
-        self.n_features_out_ = pca_result.shape[1] #needed for get_feature_names_out
+        self.n_features_out_ = pca_result.shape[1] #needed for transform
```
anneliesdej, line 220:
```suggestion
        self.n_features_out_ = np.atleast_2d(pca_result) #needed for transform
```
Just a thought this. Seems like n_components=1 could give some inconsistencies, but I did not check this because of the network issues

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,95 @@
...
+       unseen_categories = np.setdiff1d(X_val[~X_missing], list(self.mapping_.keys()))
+       if unseen_categories.size >0:
+           raise ValueError(f"transform received unseen categories. Unseen values: {unseen_categories}. Ensure input was fitted")
+
+    def _apply_mapping(self,X_val):
+
+        if X_val.shape[0] == 0:
+            return np.empty(shape= (0,self.n_features_out_),dtype=np.float32)
+ 
+        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32)#preallocation
```
anneliesdej:
```suggestion
        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32) #pre-allocation
```
British English

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,95 @@
...
+        This helper converts input to a 1-dimensional array while allowing both:
+          - 1D arrays of shape (n_samples,)
+          - 2D single-column arrays of shape (n_samples, 1)
+
+
+        Arrays with more than one column are rejected.
+        """
+        if arr.ndim >1:
+            if arr.shape[1] !=1:
+                raise ValueError("input should be 1D or be 2D with one column")
```
anneliesdej:
```suggestion
                raise ValueError(f"Expected a 1D or a 2D array with exactly one column. Received shape {arr.shape}.")
```

anneliesdej (changes requested):

src/synthpop/data_processing/encoders.py
KoomenErina: `result` is a 2D array. Setting a row with something that is not an array might give the impression that `result` is a 1D array.

src/synthpop/data_processing/encoders.py
KoomenErina: What is the unexpected behaviour? Do you have docs/example code?
( I cannot directly apply the suggestion due to invalid syntax)

anneliesdej: While looking at more sources, it might have been updated already for newer numpy. Given that poetry install did not work I did not have the correct numpy version. So the current implementation seems to be technically correct. However, a more explicit construction using the np.array may improve readability and make the intended output shape and dtype clearer to future maintainers. It is explicitly 1D now whereas the code before is 1D because valid_targets happens to be 1D

src/synthpop/data_processing/encoders.py
anneliesdej: If this gives the same result, I would prefre atleast_2d as it is clearer

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,99 @@
...
+        
+        """
+        Transform all missing values in a string array to np.nan.
+        :param x: an array of strings
+        :return: the 1-dimensional array of strings with one value for missing
+        """
+        arr = to_stringdtype_array(x)
+        return self.to_1D(arr)
+
+    def _check_unseen_values(self,X_val):
```
anneliesdej, line 50: The code block in this function (all lines) is not properly indented

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,99 @@
...
+       X_missing = np.isnan(X_val)
+        
+       # Detect unseen categories
+       seen = set(self.mapping_.keys())
+observed = set(X_val[~X_missing])
+unseen = observed - seen
+       if unseen_categories.size >0:
```
anneliesdej:
```suggestion
       if unseen:
```
unseen_categories no longer exists

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,99 @@
...
+           raise ValueError(f"transform received unseen categories. Unseen values: {unseen_categories}. Ensure input was fitted")
```
anneliesdej:
```suggestion
           raise ValueError(f"transform received categories that were not observed during fitting. Unseen values: {sorted(unseen)}. Ensure input was fitted")
```

src/synthpop/data_processing/encoders.py
```diff
@@ -107,30 +171,22 @@ def fit(self,X:npt.ArrayLike, y: npt.ArrayLike) -> Self:
...
+            raise ValueError("Number of observations in X and y do not match")
 
         # the core of this implementation
 
 
         #The alternative to using pandas here is either use scipy or DIY.
         missing_contingency_table = pd.crosstab(X_val,pd.isna(y_val))
-        if not False in missing_contingency_table:
+
+        if (missing_contingency_table.columns).all():# If there are only missing values for y:
```
anneliesdej:
```suggestion
        if np.isnan(y_val).all():# If there are only missing values for y:
```
This code matches the comment clearer. pd.isna could also be used if that seems safer

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,99 @@
...
+from synthpop.utils import to_stringdtype_array
 
 
-class PCAEncoder(TransformerMixin, BaseEstimator):
+
+
+class _BaseEncoder(TransformerMixin, BaseEstimator):
+    def to_1D(self,arr):
+        """
+        The expected input of the encoders is 2D.
```
anneliesdej:
```suggestion
        The expected input of the encoders at run-time is 2D.
```
Does this make a clearer distinction?

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,99 @@
...
+        Conceptually, the encoders operate on a single feature column at a time.
+        This helper converts input to a 1-dimensional array while allowing both:
```
anneliesdej:
```suggestion
        This helper internally converts the 2D input to a 1-dimensional array while allowing both:
```

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,99 @@
...
+            return result.reshape((-1,1))
+
+        return result
+        
+
+class PCAEncoder(_BaseEncoder):
     """
     Transforms categorical data to one or more numeric columns.
-    The user can adjust the amount of principle components by passing an instance of sklearn.decomposition.PCA to `pca_transform`
+    The user can adjust the amount of principal components by passing an instance of sklearn.decomposition.PCA to `pca_transform`
```
anneliesdej:
```suggestion
    The user can adjust the number of principal components by passing an instance of sklearn.decomposition.PCA to `pca_transform`
```
British

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,99 @@
...
+                raise ValueError(f"Expected a 1D or a 2D array with exactly one column. Received shape {arr.shape}.")
+            else:
+                return arr.reshape(-1)
+            
+        return arr
+
+    def validate_string_array(self,x):
+        
+        """
+        Transform all missing values in a string array to np.nan.
```
anneliesdej: This is under-specified. It also converts dtype, reshapes and validates dimensionality. It has a bigger role than the current description

src/synthpop/data_processing/encoders.py
```diff
@@ -4,23 +4,99 @@
...
+ 
+        result = np.full((len(X_val),self.n_features_out_), np.nan, dtype=np.float32) #pre-allocation
+
+        X_missing = np.isnan(X_val)
+
+        if X_missing.all():
+            return result
+
+        unique_vals,rev_index = np.unique(X_val,return_inverse=True)
+        #Note that rev_index does not reconstruct X_val when X_val is a stringDType array with np.nan:
```
anneliesdej, line 73: This explanation is excellent

src/synthpop/data_processing/encoders.py
anneliesdej: That is correct. However, the right-hand side visually looks like a single scalarish replacement so you have to mentally simulate np rules. Maybe a clearer alternative that shows both would then be `result[X_missing, :] = np.nan`? This explicitly shows the 2D structure as well as the selected rows + all columns

anneliesdej (changes requested):

src/synthpop/data_processing/encoders.py
KoomenErina: I assume the suggested code is intended to be
```
self.mapping_[cat] = np.array([np.mean(valid_targets)], dtype=np.float32)
```

The default behaviour of `np.array` is to make a copy. I assume that `valid_targets` has the default numpy float type (float64). So this code would make a cast and a copy. If `valid_targets` has a different shape, then the result of `np.mean(valid_targets)` will have another shape too. So it still depends on `valid_targets` being 1D. 

I think that the impact of this is small, but I am slightly in favor of the marignal performance benefit.

anneliesdej: Yes good point. Agreed on the benefit

anneliesdej (approved):

PR #108, opened 2026-05-18
102 update tree methods to use stringdtype

KoomenErina:
The end goal is to have a tree method that pass all integration tests.

Edit (scope):
- only changes of the tree methods, changes in other parts should be other issues/pull requests.
- changing StringDType to standard numpy string dtypes should happen within the tree methods and not in the missing handling. 
- changed integration tests are in scope of this pull request


KoomenErina:
The integration tests of the tree method pass now, but in order to do so I had to remove some validation checks and adjust some dtypes.

`DecisionTreeClassifier` from sklearn cannot have `string_dtype` as input for the target. So I adjusted `ReplaceMissingWithValue.prepare_data_for_fit` to return the standard fixed-length dtype for string.

When synthesising, the result from `sample_from_leaves` is again a fixed-length string. But then the validation in `ReplaceNoneWithValue.post_synth_transform` raises an exception because it is not StringDType. 

and also: `ReplaceNoneWithValue` needs to properly support dictionaries as input.

src/synthpop/methods/cart_synth.py
```diff
@@ -1,5 +1,5 @@
 """
-This module contains the CART method for synthesising data. 
+This module contains the CART method for synthesising data. Test.
```
anneliesdej:
```suggestion
This module contains the CART method for synthesising data.
```

src/synthpop/methods/cart_synth.py
```diff
@@ -17,6 +17,9 @@
 import synthpop.methods.tree_utils as tree_utils
 from synthpop.utils import validate_y, validate_dict_x
 
+def to_fixed_lenght_string_array(a):
```
anneliesdej:
```suggestion
def to_fixed_length_string_array(a):
```
Needs to be updated on every occurrence

src/synthpop/methods/cart_synth.py
```diff
@@ -50,6 +53,10 @@ def _new_tree_sampler(self):
     def _new_tree(self):
         return clone(self.tree) if self.tree is not None else self._get_tree()
 
+    def _convert_y(self,y):
```
anneliesdej, line 60: If this is only a function for the classifier, I would make it a class function there

src/synthpop/methods/cart_synth.py
anneliesdej: Should be updated to the universal helper validate_2d_dict and validate_1d_target

anneliesdej, line 167: Documentation about the data change is necessary

anneliesdej: I think output should be StringDType again, otherwise you have to do more data transformations in other classes

anneliesdej, line 131:
```suggestion

check_is_fitted(self, "target_name_")

```
Right now the function can access the unfitted attribute which will throw an AttributeError instead of NotFittedError

anneliesdej:
```suggestion
                        ,["tree_", "encoders_",  "missing_handler_", "tree_sampler_", "feature_order_"]
```
Transform depends on self.encoders_ but it is excluded in the fitted check. Also updated the order to match the parameter order (not an issue but I was writing a suggestion anyway :) )

tests/unit/test_tree_methods.py
```diff
@@ -171,7 +173,7 @@ def get_input_test_data():
         "num_1":np.array([1,2,5,2,5,3]),
         "cat_1":["a","b","c","d","e","f"],
```
anneliesdej:
```suggestion
        "cat_1":np.array(["a","b","c","d","e","f"], dtype=str_dtype),
```

tests/unit/test_tree_methods.py
```diff
@@ -171,7 +173,7 @@ def get_input_test_data():
         "num_1":np.array([1,2,5,2,5,3]),
```
anneliesdej: The methods do not seem to be tested with missing values.

tests/unit/test_tree_methods.py
```diff
@@ -358,6 +360,27 @@ def test_fit_set_feature_names_out_no_target_name(X,y,index_cat,tree_method):
 
     assert tree_method.target_name_ is None
 
+def test_fit_classifier_convert_to_str(encoder,leafnode_sampler,mocker):
```
anneliesdej:
```suggestion
def test_fit_classifier_converts_to_str(encoder,leafnode_sampler,mocker):
```

tests/integration/test_int_tree_method.py
anneliesdej, line 1:
```suggestion
```
is unused

tests/integration/test_int_tree_method.py
```diff
@@ -187,11 +150,76 @@ def test_input_to_tree_is_array_of_float32(method,X,y):
     assert isinstance(method.tree_.apply_X,np.ndarray)
     assert method.tree_.apply_X.dtype == np.dtype(np.float32)
 
+
+@pytest.mark.parametrize("method,X,y",[
```
anneliesdej: It would be better to do something like
```python
CLASSIFIER_CASES = [
get_test_data_classifier(),
get_test_data_classifier(with_cats = True),
...
]
```

and then `@pytest.mark.parametrize("X,y", CLASSIFIER_CASES)` instead of the massive parameter lists every time

tests/integration/test_int_tree_method.py
anneliesdej: This feels a little hard to reason about. Composition or subclassing would be preferred. Like `class SpyDecisionTreeClassifier(DecisionTreeClassifier)`

tests/integration/test_int_tree_method.py
```diff
@@ -35,76 +37,30 @@ def test_treemethod_regressor_fit_and_transform():
...
-
-
 def make_data_missing(X):
 
     for ik, k in enumerate(X.keys()):
-
-
-        X[k] = np.array([v if (i )% (len(X.keys())-ik) !=1 else np.nan for i,v in enumerate(X[k])],dtype=np.dtypes.StringDType(na_object=np.nan))
+        if pd.api.types.is_numeric_dtype(X[k].dtype):
+            X[k]=np.array([v if (i )% (len(X.keys())-ik) !=1 else np.nan for i,v in enumerate(X[k])])
```
anneliesdej: This is very dense and hard to read/scan

tests/integration/test_int_tree_method.py
anneliesdej:
```suggestion
            X[idx] = np.array([string.ascii_lowercase[i] for i in x_i],dtype = str_dtype)
```

tests/integration/test_int_tree_method.py
```diff
@@ -35,76 +37,30 @@ def test_treemethod_regressor_fit_and_transform():
...
+        else:
+            X[k] = np.array([v if (i )% (len(X.keys())-ik) !=1 else np.nan for i,v in enumerate(X[k])],dtype=np.dtypes.StringDType(na_object=np.nan))
```
anneliesdej:
```suggestion
            X[k] = np.array([v if (i )% (len(X.keys())-ik) !=1 else np.nan for i,v in enumerate(X[k])],dtype=str_dtype)
```

tests/integration/test_int_tree_method.py
```diff
@@ -187,11 +150,76 @@ def test_input_to_tree_is_array_of_float32(method,X,y):
...
+    assert list(X_different_order.keys()) != list(X.keys()), "test invalid, order of features has not been changed"
+
+    method.transform(X_different_order)
+
+    feature_matrix_different_order = method.tree_.apply_X
+
+    assert np.array_equal(feature_matrix_fit,feature_matrix_different_order,equal_nan=True)
+
+
 def histogram_matches(a,b):
```
anneliesdej: This name is misleading as it really checks if multiset frequency structure is preserved but not histogram equality

tests/integration/test_int_tree_method.py
```diff
@@ -9,6 +9,7 @@
 
 from synthpop.data_processing.encoders import PCAEncoder
 from synthpop.methods.cart_synth import TreeClassifierMethod, TreeRegressorMethod
+from synthpop.utils import str_dtype
 from sklearn.datasets import make_classification, make_regression
```
anneliesdej, line 14: I miss tests that explicitly check if output dtypes are correct and if missing markers survive the round-trip

anneliesdej (changes requested):

tests/unit/test_tree_methods.py
KoomenErina: The tree methods have no logic for the data, the only thing it does is delegating. 
All logic and requirements concerning missing are in the classes where the tree methods delegate to.

tests/integration/test_int_tree_method.py
KoomenErina: Composition is not preferred in this situation because the spy needs all the methods of the decision tree. Doing something like `class SpyDecisionTreeClassifier(DecisionTreeClassifier)` would introduce some duplicate code but no other significant changes. Would that make the code easier to reason about?

src/synthpop/methods/cart_synth.py
```diff
@@ -15,8 +15,12 @@
 from synthpop.methods import base_synth
 from synthpop.methods.tree_utils import LeafNodeSampler
 import synthpop.methods.tree_utils as tree_utils
-from synthpop.utils import validate_y, validate_dict_x
+from synthpop import utils 
+from synthpop.utils import validate_dict_x
```
anneliesdej:
```suggestion
```
unused and double

src/synthpop/methods/cart_synth.py
```diff
@@ -61,8 +69,8 @@ def fit(self, X: dict[str, npt.ArrayLike], y: npt.ArrayLike) -> Self:
         """
 
         self.target_name_ = getattr(y, "name", None)
-        X_val, n_samples = validate_dict_x(X)
-        y = validate_y(y, n_samples)
+        X_val, n_samples = utils.validate_2d_dict(X)
+        y = utils.validate_y(y, n_samples)
```
anneliesdej:
```suggestion
        y = utils.validate_1d_target(y, n_samples)
```
Then both unused util functions can be deleted from utils

tests/unit/test_tree_methods.py
anneliesdej: I do think it is better to use a missing example, even if it is to show that missingness is supposed to be supported when nothing really happens

tests/integration/test_int_tree_method.py
anneliesdej: I think so, yes

src/synthpop/methods/cart_synth.py
```diff
@@ -15,9 +15,10 @@
 from synthpop.methods import base_synth
 from synthpop.methods.tree_utils import LeafNodeSampler
 import synthpop.methods.tree_utils as tree_utils
-from synthpop.utils import validate_y, validate_dict_x
+from synthpop import utils 
+from synthpop.utils import validate_dict_x
 
-def to_fixed_lenght_string_array(a):
+def to_fixed_length_string_array(a):
```
anneliesdej, line 20: I would add a comment or doc string that this function explicitly does not support missing values because it is only used after ReplaceNoneWithValue. That also shows why it is not tested for missing values

src/synthpop/methods/cart_synth.py
```diff
@@ -252,6 +250,12 @@ def _get_tree(self):
         return DecisionTreeRegressor(min_samples_leaf=5, #equivalent to minbucket in synthpop-r
                                       min_impurity_decrease= 1e-08# equivalent to cp in synthpop-r
                                       ,)
+    
+    def _convert_y(self,y):
+        return y.astype(np.float32,copy=False)
+    
+    def transform(self, X):
```
anneliesdej: Should transform also be added to TreeClassifierMethod to explicitly cast as the stringdtype? Of course, it already is stringdtype because of the other classes it delegates to

tests/integration/test_int_tree_method.py
```diff
@@ -212,38 +242,32 @@ def test_order_of_input_dict_does_not_change_output(method,X,y):
...
-    (rigged_tree_classifier_method(),*get_test_data_classifier()),
-    (rigged_tree_classifier_method(),*get_test_data_classifier(with_cats=True)),
-    (rigged_tree_regressor_method(),*get_test_data_regressor()),
-    (rigged_tree_regressor_method(),*get_test_data_regressor(with_cats=True)),
-                                       ])
+    return np.array_equal(hist_a,hist_b) and (pd.isna(a).sum()==pd.isna(b).sum())
+
+
+@pytest.mark.parametrize("method,X,y",NO_MISSING_TARGET )
 def test_no_information_lost_when_fitting_tree(method,X,y):
```
anneliesdej:
```suggestion
def test_fit_preserves_feature_value_frequencies(method,X,y):
```
Better name as it does not test information preservation in general, but more the multiset preservation after encoding and build_feature_matrix

tests/integration/test_int_tree_method.py
```diff
@@ -127,20 +136,37 @@ def rigged_tree_regressor_method():
...
+    (rigged_tree_regressor_method(),*get_test_data_regressor(with_cats=True)),
     (rigged_tree_regressor_method(),*get_test_data_regressor(with_missing_features= True)),
     (rigged_tree_regressor_method(),*get_test_data_regressor(with_cats=True,with_missing_features= True)),
-    (rigged_tree_regressor_method(),*get_test_data_regressor(with_cats=True,with_missing_features= True,with_missing_target=True)),
-    (rigged_tree_classifier_method(),*get_test_data_classifier(with_cats=True,with_missing_features= True,with_missing_target=True)),
-                                       ])
+
+]
+@pytest.mark.parametrize("method,X,y",[*CLASSIFIER_CASES,*REGRESSOR_CASES ])
 def test_input_to_tree_is_array_of_float32(method,X,y):
```
anneliesdej:
```suggestion
def test_tree_received_float32_feature_matrix(method,X,y):
```
Makes the tests easier to scan

anneliesdej (changes requested):

src/synthpop/methods/cart_synth.py
KoomenErina: I don't think it would add any value.

anneliesdej: Little explanation: Classifier only works with stringDType whereas regressor could also be used with, for example, float64. So the regressor has the specific cast to float32

src/synthpop/methods/cart_synth.py
```diff
@@ -15,8 +15,15 @@
 from synthpop.methods import base_synth
 from synthpop.methods.tree_utils import LeafNodeSampler
 import synthpop.methods.tree_utils as tree_utils
-from synthpop.utils import validate_y, validate_dict_x
+from synthpop import utils 
 
+def to_fixed_length_string_array(a):
+    """
+    Converts an array of string_dtype to a array of fixed length string dtype.
```
anneliesdej:
```suggestion
    Converts an array of StringDType to a array of fixed length string dtype.
```

src/synthpop/methods/cart_synth.py
```diff
@@ -240,6 +253,12 @@ def _get_tree(self):
         return DecisionTreeRegressor(min_samples_leaf=5, #equivalent to minbucket in synthpop-r
                                       min_impurity_decrease= 1e-08# equivalent to cp in synthpop-r
                                       ,)
+    
+    def _convert_y(self,y):
+        return y.astype(np.float32,copy=False)
+    
+    def transform(self, X):
```
anneliesdej:
```suggestion
    def transform(self, X: Dict[str, npt.NDArray]) -> npt.NDArray:
```

src/synthpop/methods/cart_synth.py
```diff
@@ -50,6 +57,10 @@ def _new_tree_sampler(self):
     def _new_tree(self):
         return clone(self.tree) if self.tree is not None else self._get_tree()
 
+    def _convert_y(self,y):
+        #overwritten in TreeClassifiermethod and TreeRegressorMethod
+        return y
+
 
     def fit(self, X: dict[str, npt.ArrayLike], y: npt.ArrayLike) -> Self:
```
anneliesdej:
```suggestion
    def fit(self, X: Dict[str, npt.NDArray], y: npt.NDArray) -> Self:
```

src/synthpop/methods/cart_synth.py
```diff
@@ -1,5 +1,5 @@
 """
-This module contains the CART method for synthesising data. 
+This module contains the CART method for synthesising data.
 """
 from abc import abstractmethod, ABCMeta
 from typing import Self
```
anneliesdej:
```suggestion
from typing import Self, Dict
```

src/synthpop/methods/cart_synth.py
anneliesdej:
```suggestion
    def transform(self, X: Dict[str, npt.NDArray]) -> npt.NDArray:
```

src/synthpop/methods/cart_synth.py
anneliesdej: However, I would make the casting explicit so the output of this class is not dependent on which sampler or missing handler is used. Especially given that the doc string show that this function should only output StringDType

src/synthpop/methods/cart_synth.py
```diff
@@ -1,5 +1,5 @@
 """
-This module contains the CART method for synthesising data. 
+This module contains the CART method for synthesising data.
```
anneliesdej, line 2: The old util functions can be removed

anneliesdej (changes requested):

src/synthpop/methods/cart_synth.py
```diff
@@ -19,7 +19,7 @@
 
 def to_fixed_length_string_array(a):
     """
-    Converts an array of string_dtype to a array of fixed length string dtype.
+    Converts an array of StringDType to a array of fixed length string dtype.
```
anneliesdej:
```suggestion
    Converts an array of StringDType to an array of fixed length string dtype.
```

src/synthpop/methods/cart_synth.py
```diff
@@ -166,21 +171,26 @@ class TreeClassifierMethod(_AbstractTreeMethod):
     :param missing_handler: handler for missing values in the target variable. Default is :class:`~synthpop.data_processing.missing_value_handling.ReplaceNoneWithValue`
     :param tree_sampler: a  :py:class:`~synthpop.methods.tree_utils.LeafNodeSampler` object to sample from the leaves of the decision tree.
 
+    The output wil always be a numpy array. Numeric output will always have np.float32 as dtype. Categorical output will always have `np.dtypes.StringDType(na_object=np.nan)` as dtype.
```
anneliesdej: I think a part of this should be written at TreeRegressorMethod

src/synthpop/methods/cart_synth.py
```diff
@@ -207,22 +223,22 @@ class TreeRegressorMethod(_AbstractTreeMethod):
...
         >>> from synthpop.methods.cart_synth import TreeRegressorMethod
         >>> import numpy as np
-        >>> tree_method = TreeRegressorMethod()
-        >>> X = {
-        ...         "column1":np.array([1.1,2.2]),
-        ...         "column2":np.array([1.4,1.2]),
-        ...         "column3":np.array(["a","b"])
-        ...         }
+        >>> from synthpop.utils import str_dtype
+        >>> X = {        "column1":np.array([1.1,2.2]),        "column2":np.array([1.4,1.2]),        "column3":np.array(["a","b"],dtype=str_dtype)        }
```
anneliesdej: I can't check the readthedocs now, but this seems to look a bit odd as if it is supposed to be spread over multiple lines

src/synthpop/methods/cart_synth.py
```diff
@@ -207,22 +223,22 @@ class TreeRegressorMethod(_AbstractTreeMethod):
     :param missing_handler: handler for missing values in the target variable. Default is :class:`~synthpop.data_processing.missing_value_handling.MissingValuePredictor`
     :param tree_sampler: a  :py:class:`~synthpop.methods.tree_utils.LeafNodeSampler` object to sample from the leaves of the decision tree.
 
+    The output wil always be a numpy array. Numeric output will always have np.float32 as dtype. Categorical output will always have `np.dtypes.StringDType(na_object=np.nan)` as dtype.
```
anneliesdej: Wouldn't it be better to only mention the float output here and only the stringdtype in the classifier? Mentioning it like this can be confusing as there is only numeric output here and only categorical in the classifier

anneliesdej (changes requested):

src/synthpop/methods/cart_synth.py
anneliesdej: Here it also needs to be updated to only mention the one possible output

anneliesdej (changes requested):

anneliesdej (approved):

PR #111, opened 2026-05-19
90 implement cart method

KoomenErina:
Requirements:

- input can be "dirty" pandas dataframes/series
- input numpy arrays supported?
- implements set_output api
- passes only (dicts of) numpy arrays with correct dtypes to its dependencies
- selects correct tree method

- tune_cart method

Approach:
- one function that "cleans" data 
- assert that CartMethod uses that function and delegates correctly.


KoomenErina:
This will be taken over by @anneliesdej

Issue #114, opened 2026-05-21
Improve ReplaceMissingWithValue


KoomenErina:
`ReplaceMissingWithValue` 

Goals:
- `ReplaceMissingWithValue` should be useable for any dtype of the target, instead of only for string dtypes. 
- It should accept a dictionary for `X`

PR #115, opened 2026-05-22
114 improve replacemissingwithvalue

KoomenErina:
see #114


src/synthpop/data_processing/missing_value_handling.py
anneliesdej, line 266: Should we rename it to ReplaceMissingWithValue?

anneliesdej (changes requested):
All new code changes look good, however the integration test is not updated yet and we need to consider renaming the class

tests/unit/test_replace_missing_with_value.py
anneliesdej, line 1: unit tests seem good, integration test is not updated yet

anneliesdej (approved):

PR #116, opened 2026-05-26
81 implement pairwise spmse

casvanerpduo:
Ik heb een eerste versie van de S_pSME geschreven. 

Specifieke dingen waar t goed is om naar te kijken zijn:
1. De gehele structuur van de code/tests. Veel is nog nieuw dus wellicht staan dingen niet logisch uitgewerkt en/of stukken 'testbare code' voldoende losgekoppeld. Ook maak ik veel gebruik van dictionaires om de gematchte variabelen X en Y in op te slaan als {(kolom1, kolom2): [.....]}. Dit vereist extra regels code, wat wellicht in pandas zelf al efficiënter kan.
2. Met name regel 82-87 kunnen misschien eleganter/robuuster worden opgeschreven (bv. als dictionaires kunnen worden vermeden).
3. de functie Calc_S_pSME: Ik denk dat deze ook beter geschreven kan worden met in ieder geval minder functie argumenten.
4. de input tests (die nu allemaal een ValueError geven als fout) kunnen wellicht in een eigen functie.


pyproject.toml
```diff
@@ -35,6 +35,11 @@ dev = [
 [tool.poetry]
 packages = [{include = "synthpop", from = "src"}]
 
+[[tool.poetry.source]]
```
KoomenErina: this source does not work on everyones pc

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,99 @@
...
+    "df_input, output",
+    [
+    (pd.DataFrame({"c1" : [0,0]}), {('c1', 'c1'): pd.DataFrame({"c1" : [0], 'x': [2]})}),
+    (pd.DataFrame({"c1" : [0,0], "c2" : [1,1]}), {('c1', 'c1'): pd.DataFrame({"c1" : [0], 'x': [2]}),
+                                                    ('c1', 'c2'): pd.DataFrame({"c1" : [0], "c2" : [1], 'x': [2]}),
+                                                    ('c2', 'c2'): pd.DataFrame({"c2" : [1], 'x': [2]})})
+    ] 
+)
+def test_joint_frequencies(df_input,output):
+    res = joint_frequencies(df_input, 'x')
```
KoomenErina: why the second parameter?

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,99 @@
...
+
+def test_preprocessing():
+    df = pd.DataFrame({
+        "a": [0, 25, 50, 75, 100],
+        "b": ["this", "is", "a", "test", "case"]
+    })
+    num_bins=10
+    
+    result = preprocessing(df.copy(), max_bins=num_bins)
+    expected_a = pd.cut(df["a"], num_bins, labels=range(num_bins)) #run the pd cut on numeric column
```
KoomenErina: When you use `pd.cut` here, you are doing a calculation in the unit test. Furthermore, the test depends on the behaviour of pandas.  

One problem with this is that it is not clear what the expected bins are. This makes it harder to understand the indended behaviour of the code. 

Second, it weakens the this unit test. If `pd.cut` is misunderstood, it will be misunderstood in both the unit test and the code, causing the test to pass eventhough it does not do what is intended. 

It is better to manually define the bins for this numeric example.

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,99 @@
...
+        ('c2', 'c2'): pd.DataFrame({"c2" : [1,2], 'f_syn': [2,2]})},
+        False) #In this case f_orig=f_synth, so outcome statistic should be 0!
+
+    ] 
+)
+def test_calc_S_pMSE(orig_df, syn_df, should_raise):
+    n_s, n_o = 2, 2
+    f_origxy= 'f_or'
+    f_synthxy= 'f_syn'
+    if should_raise:
```
KoomenErina: A unit test should not have such a branching. It is better to have separete test for the cases it should raise an error. That allows you to assert the error message aswell.

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,99 @@
...
+    ] 
+)
+def test_joint_frequencies(df_input,output):
+    res = joint_frequencies(df_input, 'x')
+    for k in output:
+        assert k in res
+        assert res[k].equals(output[k])
+
+@pytest.mark.parametrize(
+    "orig_df, syn_df, should_raise",
```
KoomenErina: The names `orig_df` and `syn_df` are usually used for the observed data and the synthetic data.
It seems like that is not the case in this scenario, as `orig_df` seems to refer to the joined frequency in this test.

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,99 @@
...
+    f_synthxy= 'f_syn'
+    if should_raise:
+        with pytest.raises(ValueError):
+            for key in orig_df:
+                Calc_S_pSME(orig_df[key], syn_df[key], n_o, n_s,  f_origxy, f_synthxy)
+    else:
+        res = {}
+        for key in orig_df:
+            res[key] = Calc_S_pSME(orig_df[key], syn_df[key], n_o, n_s,  f_origxy, f_synthxy)
+        assert all(v == 0. for v in res.values())
```
KoomenErina: It seems like all test cases that don't raise an exception have an result identically zero. Some testcases with non-zero result are needed too.

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,99 @@
...
+    "orig_df, syn_df, expected",
+    [
+        (pd.DataFrame([[1,2],[3,4]], columns=['c1', 'c2']), 
+        pd.DataFrame([[1,2],[3,4]], columns=['c1', 'c2']), 
+        pd.DataFrame([['c1', 'c1', 0],['c1', 'c2', 0],['c2','c2',0]], columns=['column1', 'column2', 'S_pSME'])) #Desired output format.
+    ] #The S_pSME should be 0 for all three as the synthetic and original are the exact same!
+)
+def test_pairwise_spmse_output(orig_df, syn_df, expected):
+
+    output = pairwise_spmse(orig_df, syn_df)
```
KoomenErina: `pairwise_spmse` calls all other functions. 
If you add a case where the S_pSME is not zero, you can remove many of the other tests.

KoomenErina:
It might be helpfull to consider [`pd.Series` ](https://pandas.pydata.org/docs/reference/api/pandas.Series.html) for the joined frequencies. That removes the need for arbitrary column names.

KoomenErina (changes requested):

tests/unit/test_S_pMSE.py
casvanerpduo: As a label name for the joint frequency table.

I removed the .reset_index() function such that a column name for the joint frequency table is not required.

tests/unit/test_S_pMSE.py
casvanerpduo: I changed the structure of my unit tests, where I now solely test the entirety of  the S_pSME function (including all subfunctions) on Input/ValueErrors, Warnings, as input to output.

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.2.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.4.1 and should not be changed by hand.
```
KoomenErina: This file should be removed from this pull request

pyproject.toml
```diff
@@ -43,4 +43,4 @@ build-backend = "poetry.core.masonry.api"
 pythonpath = "src"
```
KoomenErina, line 43: This file should be removed from this pull request

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,96 @@
...
+            df.loc[mask, colname] = na_label
+    return df
+
+def joint_frequencies(df):
+    """
+    Create pairwise joint frequency tables for all variable pairs.
+    """
+    results = {}
+    for col1, col2 in combinations_with_replacement(df.columns, 2):
+        if col1 == col2:
```
KoomenErina: perhaps the content of this loop can be simplified using `pd.crosstab`

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,88 @@
...
+
+@pytest.mark.parametrize(
+    "orig_df, syn_df, expected, na_label",
+    [
+        (pd.DataFrame([[1,2],[3,4]], columns=['c1', 'c2']), 
+        pd.DataFrame([[1,2],[3,4]], columns=['c1', 'c2']), 
+        pd.DataFrame([['c1', 'c1', 0.],['c1', 'c2', 0.],['c2','c2',0.]], columns=['column1', 'column2', 'S_pMSE']), 'N.a.N.'),#Desired output format.
+        #Results should all be zero as original dataset=synthetic
+
+        (pd.DataFrame([['a', 0], ['a', 0], ['b',1]], columns=['c1', 'c2']),
```
KoomenErina: I find it difficult to visualize dataframes written like this. Consider using a dictionary to define the dataframes, so that it is immediately clear which data is in which column.

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,88 @@
...
+        (pd.DataFrame({"A": [10], "B": [20]}), pd.DataFrame({"A": [10], "C": [20]}), 25), #Check column names not equal
+        ([],[], 12), #Check for non pandas dataframes
+        (pd.DataFrame(), pd.DataFrame(), 25.), #Check if max_bins is not an integer
+        (pd.DataFrame(), pd.DataFrame(), -12), #Check for negative bins
+        (pd.DataFrame(), pd.DataFrame(), 35), #Check empty DataFrames
+        (pd.DataFrame(['a', 'N.a.N.'], dtype=str), (pd.DataFrame(['a', 'N.a.N.'], dtype=str)), 25) #Check if N.a.N. is already in use
+    ] 
+)
+def test_pairwise_spmse_inputtests(orig_df, syn_df, max_bins):
+    with pytest.raises(ValueError):
```
KoomenErina: when asserting that an error is raised, it's best to also test that the correct error message is included. This ensures that the exception is raised for the intended reason.

KoomenErina (changes requested):
The following tests are missing:
- test with various values of `max_bin`
- multiple occurences of "missing"
- column with string dtype
- column with explicit categorical dtype
- test with only missing values.

Try to use 1 for-loop over the pairs of columns, and calculate everything per column pair. This will probably simplify your data structures and make the code easier to understand.

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,129 @@
...
+        #S_pMSE should all be zero as the original_dataset=the synthetic_dataset
+        
+        (pd.DataFrame({"c1": ["a", "a", "b"],"c2": [0, 0, 1]}),
+        pd.DataFrame({"c1": ["a", "b", "b"],"c2": [1, 0, 1]}),
+        pd.DataFrame({"column1": ["c1", "c1", "c2"],"column2": ["c1", "c2", "c2"],"S_pMSE": [2/3, 2/3, 2/3]}),"N.a.N.", 1000), 
+        #A non-zero answer to the S_pMSE, calculated by hand, with a high value of bins
+        
+        (pd.DataFrame({"c1": ["a", "a", "b"],"c2": [0, 0, 1]}),
+        pd.DataFrame({"c1": ["a", "b", "b"],"c2": [1, 0, 1]}),
+        pd.DataFrame({"column1": ["c1", "c1", "c2"],"column2": ["c1", "c2", "c2"],"S_pMSE": [2/3, 2/3, 2/3]}),"N.a.N.", 2),
```
KoomenErina: There is no test that shows that `max_bins` has effect on the s_pmse. I could "hard code" max_bins to 10 without causing any test to fail. A test where the `max_bins` causes the s_psme to be different is needed.

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,129 @@
...
+        pd.DataFrame({"column1": ["c1", "c1", "c2"],"column2": ["c1", "c2", "c2"],"S_pMSE": [2/3, 2/3, 2/3]}),"N.a.N.", 2), 
+        #Check for two bins
+
+        (pd.DataFrame({"c1": [0, 0, 0, 1]}), 
+        pd.DataFrame({"c1": [0, 1]}),
+        pd.DataFrame({"column1": ["c1"], "column2": ["c1"], "S_pMSE": [1/4]}), "N.a.N.", 25),
+        #A one-dimensional input with different number of rows
+
+        (pd.DataFrame({"c1": [0, 0, 0, 1]}, dtype='category'), 
+        pd.DataFrame({"c1": [0, 1]}, dtype='category'),
```
KoomenErina: categorical data is usually used with string labels. There should be a test that uses that. See https://pandas.pydata.org/docs/user_guide/categorical.html

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,129 @@
...
+
+    ] #Data where every value will fall into the same bin
+)
+def test_pairwise_spmse_output(orig_df: pd.DataFrame, syn_df: pd.DataFrame, expected: pd.DataFrame, na_label : str, max_bins : int):
+
+    output = pairwise_spmse(orig_df, syn_df, na_label=na_label, max_bins=max_bins)
+    assert output.equals(expected)
+
+
+@pytest.mark.parametrize(
```
KoomenErina, line 183: This test is parametrized, but has only one set of data. Parametrizing is not needed here

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,147 @@
...
+    :type: DataFrame
+    :param col1: column name 1
+    :type: str
+    :param col1: column name 2
+    :type: str
+    """
+
+    if col1 == col2:
+        all_idx = df[col1].unique()
+        jf = (df.groupby(col1, dropna=False).size().reindex(all_idx, fill_value=0))
```
KoomenErina: perhaps `value_counts` is more compact? https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.value_counts.html#pandas.DataFrame.value_counts

`DataFrame.size` is not intended for this usecase

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,147 @@
...
+        raise ValueError("Original and synthetic dataframes must have the same shape and column names.")
+    
+    if max_bins < 1 or not isinstance(max_bins, int):
+        raise ValueError("The number of bins should be an integer with value larger than 1.")
+
+    #Start calculations for preprocessing here
+    for column_name in orig_df:
+
+        if pd.api.types.is_numeric_dtype(orig_df[column_name]):
+            combined = pd.concat([orig_df[column_name], syn_df[column_name]])
```
KoomenErina: Note that `combined` is still 1D. It appends the 2 series. Is this intentional?

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,129 @@
...
+    "orig_df, syn_df, expected, na_label, max_bins",
+    [
+
+        (pd.DataFrame({"c1": [1, 3],"c2": [2, 4]}), 
+        pd.DataFrame({"c1": [1, 3],"c2": [2, 4]}), 
+        pd.DataFrame({"column1": ["c1", "c1", "c2"], "column2": ["c1", "c2", "c2"], "S_pMSE": [0.0, 0.0, 0.0]}), 'N.a.N.', 25),
+        #S_pMSE should all be zero as the original_dataset=the synthetic_dataset
+        
+        (pd.DataFrame({"c1": ["a", "a", "b"],"c2": [0, 0, 1]}),
+        pd.DataFrame({"c1": ["a", "b", "b"],"c2": [1, 0, 1]}),
```
KoomenErina: In practice it might happen that the original dataset has a category that is not present in the synthetic dataset. This has caused problems in the past. It would be good to add a test for that.

KoomenErina (changes requested):

src/synthpop/utility_metrics/spmse.py
casvanerpduo: Yes, as the sole usecase of this combined DataFrame is to find binedges that will be the same for the synthetic and original dataframe. I concatenate to make sure that I take the global min/max values that in some cases might only be present in one of the dataframes

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
...
+    :type: str
+    :param col1: column name 2
+    :type: str
+    """
+
+    if col1 == col2:
+        all_idx = pd.Index(df_orig[col1].unique()).union(df_syn[col1].unique())
+    
+    else:
+        idx1 = pd.Index(df_orig[col1].unique()).union(df_syn[col1].unique())
```
anneliesdej: Doing this twice can be avoided to save some computation costs.

If you precompute once:
```python
all_levels = {}

for col in orig_df.columns:
    all_levels[col] = (
        pd.Index(orig_df[col].unique())
        .union(pd.Index(syn_df[col].unique()))
    )
```
and then set
```
idx1 = all_levels[col1]
idx2 = all_levels[col2]
```
it would be more optimised

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
...
+    :type: Either pd.MultiIndex or list of names
+    """
+
+    if col1 == col2:
+        jf = df[col1].value_counts(dropna=False).reindex(full_idx, fill_value=0).rename(col1)
+    else:
+        jf = df[[col1,col2]].value_counts(dropna=False).reindex(full_idx, fill_value=0).rename(col1+','+col2)
+    return jf
+
+def Calc_S_pSME(jf_or: pd.Series, jf_syn: pd.Series, n_o: int, n_s: int):
```
anneliesdej:
```suggestion
def calc_spmse(jf_or: pd.Series, jf_syn: pd.Series, n_o: int, n_s: int):
```
Better PEP8 compliance

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
...
+    """
+
+    if pd.api.types.is_numeric_dtype(column):
+
+        if column.notna().any():
+            binned_column = pd.cut(column,bins) 
+            column = binned_column
+
+    else:
+        mask = pd.isna(column)
```
anneliesdej: Functional description needs to be updated to match this behaviour, as we agreed this was a good way to handle the issue

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
...
+    n_o = len(orig_df.index)
+    n_s = len(syn_df.index)
+
+    if n_o == 0 or n_s ==0:
+        raise ValueError('Both the original and synthetic dataframe should consist out of non-zero rows')
+    
+    if len(orig_df.columns) != len(syn_df.columns) or not all(orig_df.columns==syn_df.columns):
+        raise ValueError("Original and synthetic dataframes must have the same shape and column names.")
+    
+    if max_bins < 1 or not isinstance(max_bins, int):
```
anneliesdej, line 163: The error message is not in line with what this code tests. According to the error message it should be `max_bins <= 1`

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
...
+    :param n_s: number of rows in synthetic dataset
+    :type: int
+    """
+
+    rescaled_differences = jf_syn - (n_s / n_o) * jf_or
+    expected_frequency = (jf_or + jf_syn) * n_s/ (n_o + n_s)
+
+    k = (expected_frequency>0)
+    if k.sum()>1:
+        S_pMSE = 1/(k.sum()-1) * np.nansum(rescaled_differences[k]**2) / np.nansum(expected_frequency[k])
```
anneliesdej: This is not in line with the functional description. The FD states the sum of (delta^2/e). This now creates sum(delta^2)/sum(e). Either the implementation or the functional description is incorrect

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
...
+
+    else: #k=1 
+        S_pMSE = 0.
+        warnings.warn(f'both variables are constant and equal, so the statistic is undefined. Return 0 for variable pair: {jf_syn.name}')
+
+    return S_pMSE
+
+def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int = 25) -> pd.DataFrame:
     """
     Compute the Standardized propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables
```
anneliesdej:
```suggestion
    Compute the Standardised propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables 
```
We follow British English in this package. There are more words that are still written as American English (like discretized -> discretised)

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
 This module contains metrics to evaluate the utility of synthetic data.
 """
 import pandas as pd
+from itertools import combinations_with_replacement
+import numpy as np
+import warnings
 
-def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_groups: int = 25, na_label: str = "__NA__") -> pd.DataFrame:
+def preprocessing(column: pd.Series, bins=None):
```
anneliesdej:
```suggestion
def preprocessing(column: pd.Series, bins: Sequence[float] | None =None):
```
Not really needed for an internal function, but could be added if wanted

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
...
+    """
+    Makes all the required indices of possible combinations (x,y), for x as element of col1 and y as element of col2
+    Takes both the synthetic as the original dataset into account, as specific combinations might be present in one or the other
+    :param orig_df: Original dataset.
+    :type orig_df: pd.DataFrame
+    :param syn_df: Synthetic dataset. 
+    :type syn_df: pd.DataFrame
+    :param col1: column name 1
+    :type: str
+    :param col1: column name 2
```
anneliesdej:
```suggestion
    :param col2: column name 2
```

src/synthpop/utility_metrics/spmse.py
anneliesdej:
```suggestion
def _preprocessing(column: pd.Series, bins=None):
```
Python has no true private functions, but adding a leading underscore is a well-established convention to show that a function is meant to be internal. You could add this to all functions excpet for `pairwise_spmse`

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,156 @@
 This module contains metrics to evaluate the utility of synthetic data.
 """
 import pandas as pd
+from itertools import combinations_with_replacement
+import numpy as np
+import warnings
```
anneliesdej, line 7:
```suggestion

__all__ = ["pairwise_spmse"]

```
Adding this would make sure that if someone runs from spmse import * would only import pairwise_spmse and not all other functions. This way only that function is part of the public interface

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,132 @@
+from typing import Any, Literal
+
+import pandas as pd
+import numpy as np
+import pytest
+
+from synthpop.utility_metrics.spmse import pairwise_spmse
+
```
anneliesdej, line 6: Most edge cases from the FD are covered. I would add a test that checks if column order matters, as it seems in the implementation it does but the FD specifically mentions "A synthetic dataset with $n_s$ rows and the same $p$ columns. The order of the columns is not relevant, but there needs to be a check that the column names match between both datasets."

anneliesdej, line 6: I would also add this test:
```python

def test_pairwise_spmse_does_not_mutate_inputs():
    orig = pd.DataFrame(
        {
            "numeric": [1, 2, 3, 4],
            "categorical": ["a", "b", "a", None],
        }
    )

    syn = pd.DataFrame(
        {
            "numeric": [1, 2, 4, 5],
            "categorical": ["a", "b", None, "c"],
        }
    )

    orig_before = orig.copy(deep=True)
    syn_before = syn.copy(deep=True)

    pairwise_spmse(orig, syn)

    pd.testing.assert_frame_equal(orig, orig_before)

    pd.testing.assert_frame_equal(syn, syn_before)
```

However, this test will fail with the current implementation because preprocessing modifies the dtype of the input

src/synthpop/utility_metrics/spmse.py
casvanerpduo: I changed this to value_count, but than back to groupby.size() as that better handles NaN values with inclusion of MultiIndex( https://stackoverflow.com/questions/76224122/including-zero-counts-for-unobserved-categoricals-in-pandas-dataframe-value-coun )

src/synthpop/utility_metrics/spmse.py
casvanerpduo: For this I will open a new issue

src/synthpop/utility_metrics/spmse.py
```diff
@@ -6,63 +6,54 @@
 import numpy as np
 import warnings
```
KoomenErina, line 7:
````suggestion
from typing import Sequence
````

File: tests/unit/test_S_pMSE.py [hunk 30]
```diff
@@ -39,91 +39,145 @@ def test_pairwise_spmse_inputtests(orig_df, syn_df, max_bins, error):
...
+        pd.DataFrame({"column1": ["c1"], "column2": ["c1"], "S_pMSE": [9/16]}), 25), 
         #Check missing value handling, np.nan+integers (float DataFrame)
 
         (pd.DataFrame({"c1": ['a', 'a', 'a', pd.NA]}), 
         pd.DataFrame({"c1": ['a', pd.NA]}),
-        pd.DataFrame({"column1": ["c1"], "column2": ["c1"], "S_pMSE": [1/4]}), "N.a.N.", 25), 
+        pd.DataFrame({"column1": ["c1"], "column2": ["c1"], "S_pMSE": [9/16]}), 25), 
         #Check missing value handling, (str + pd.NA)
 
         (pd.DataFrame({"c1": ['a', 'a', 'a', 'Different N.a.N.']}),
```
KoomenErina: Is this testcase still relevant?

src/synthpop/utility_metrics/spmse.py
```diff
@@ -115,34 +119,37 @@ def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int =
     if n_o == 0 or n_s ==0:
         raise ValueError('Both the original and synthetic dataframe should consist out of non-zero rows')
     
-    if len(orig_df.columns) != len(syn_df.columns) or not all(orig_df.columns==syn_df.columns):
+    if len(orig_df.columns) != len(syn_df.columns) or set(orig_df.columns) != set(syn_df.columns):
```
KoomenErina:
```suggestion
    if  orig_df.columns.equals(syn_df.columns):
```
This is more compact, and more explicit in the intention of the code

src/synthpop/utility_metrics/spmse.py
```diff
@@ -115,34 +119,37 @@ def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int =
...
         raise ValueError("Original and synthetic dataframes must have the same shape and column names.")
     
     if max_bins < 1 or not isinstance(max_bins, int):
-        raise ValueError("The number of bins should be an integer with value larger than 1.")
+        raise ValueError("The number of bins should be an integer with value of at least 1.")
+
+    #Make sure the original DataFrames are not modified 
+    o_df = orig_df.copy(deep=True)
+    s_df = syn_df.copy(deep=True)
```
KoomenErina: a deep copy is not needed. A deep copy increases memory usage.
See https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.copy.html
and https://pandas.pydata.org/docs/user_guide/copy_on_write.html#copy-on-write

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,154 @@
...
 
-def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_groups: int = 25, na_label: str = "__NA__") -> pd.DataFrame:
+__all__ = ["pairwise_spmse"]
+
+def _preprocessing_numeric(column: pd.Series, bins: Sequence[float] | None =None):
+    """
+    Preprocessing of the dataframes s.t. S_pMSE statistic can be calculated
+    Bins numerical values in bins
+    :param column: the specific column. 
+    :type: pd.DataFrame Datatypes can be numeric, categorical, or string
```
KoomenErina: writing the types in docstring is not needed. Using typehints is sufficient.

KoomenErina (changes requested):
The lack of pyproject.toml is causing problems with building the docs. I think it is best to include the version of pyproject.toml as it is in the develop branch.

docs/source/functional descriptions/S_pMSE.md
```diff
@@ -26,7 +26,7 @@ The computation of S_pMSE for a given pair of variables $(X, Y)$ (from the origi
 
 For each variable pair $(X, Y)$:
 - If a variable is numeric, it is discretised into at most $\text{max\_bins}$ bins.
-- Missing values in either variable are replaced with the value "N.a.N.". If that value is already used, an error is raised.
+- Missing values (`pandas.NA`, `NumPy.nan`, `None` ) are replaced with `NumPy.nan`$.
```
anneliesdej: I think this should be changed to something like
```missing values are all normalised to the same missing value (i.e., `NumPy.nan`).```
Then it is more independent of programming languages

File: tests/unit/test_S_pMSE.py [hunk 35]
```diff
@@ -0,0 +1,177 @@
...
+
+        (pd.DataFrame([0]), pd.DataFrame([0]), -12, "with value of at least 1."), 
+        #Check for negative bins
+
+        (pd.DataFrame(), pd.DataFrame(), 35, "dataframe should consist out of non-zero rows"), 
+        #Check empty DataFrames
+
+    ] 
+)
+def test_pairwise_spmse_inputtests(orig_df, syn_df, max_bins, error):
```
anneliesdej:
```suggestion
def test_pairwise_spmse_raises_wrong_inputs(orig_df, syn_df, max_bins, error):
```
A different name makes it clearer what is tested exactly

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
+        warnings.warn(f'both variables are constant and equal, so the statistic is undefined. Return 0 for variable pair: {jf_syn.name}')
+
+    return S_pMSE
+
+def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int = 25) -> pd.DataFrame:
     """
-    Compute the Standardized propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables 
+    Compute the Standardised propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables 
     between two similarly-structured dataframes: one original dataset and one synthetic version of the original dataset,
     assuming the same variables and the same number of rows.
```
anneliesdej:
```suggestion
    assuming the same variables. The number of rows may differ.
```
The implementation now actually supports a different number of rows.

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,177 @@
+from typing import Any, Literal
```
anneliesdej: The following tests currently fail:
=========================== short test summary info ===========================
FAILED tests/unit/test_S_pMSE.py::test_pairwise_spmse_output[orig_df1-syn_df1-expected1-1000]
FAILED tests/unit/test_S_pMSE.py::test_pairwise_spmse_output[orig_df2-syn_df2-expected2-25]
FAILED tests/unit/test_S_pMSE.py::test_pairwise_spmse_output[orig_df6-syn_df6-expected6-25]
FAILED tests/unit/test_S_pMSE.py::test_pairwise_spmse_output[orig_df10-syn_df10-expected10-25]
FAILED tests/unit/test_S_pMSE.py::test_pairwise_spmse_output[orig_df15-syn_df15-expected15-3]
FAILED tests/unit/test_S_pMSE.py::test_warning - Failed: DID NOT WARN. No war...


**Edit**: issue from my side. I did not have a new enough pandas version

anneliesdej:
```suggestion
```
These imports are not used

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,177 @@
...
+import pytest
+
+from synthpop.utility_metrics.spmse import pairwise_spmse
+
+@pytest.mark.parametrize(
+    "orig_df, syn_df, max_bins, error",
+    [
+        
+        (pd.DataFrame([[1,2],[3,4]]), pd.DataFrame([[1,2,3],[4,5,6]]), 25, "must have the same shape and column names."), 
+        #Check for unequal number of columns
```
anneliesdej:
```suggestion
        # Check for unequal number of columns
```
Try to take a look at the issue about PEP8 (#122 ). For example, it requires a space after the `#` for a comment

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
+
+        S_pMSE = 0.
+        warnings.warn(f'both variables are constant and equal, so the statistic is undefined. Return 0 for variable pair: {jf_syn.name}')
+
+    return S_pMSE
+
+def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int = 25) -> pd.DataFrame:
     """
-    Compute the Standardized propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables 
+    Compute the Standardised propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables
```
anneliesdej: This probably used to be taken from the FD, but the reference does not exist in the docstring

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
+    
     :param orig_df: Original dataset.
     :type orig_df: pd.DataFrame
     :param syn_df: Synthetic dataset. Should have the same columns as the original dataset, in the same order and the same number of rows.
     :type syn_df: pd.DataFrame
-    :param max_groups: Maximum number categories in which numeric variables can be discretized. Default value is 25.
-    :type max_groups: int
-    :param na_label: String value to be given to missing values. Default is __NA__
-    :type na_label: str
+    :param max_bins: Maximum number categories in which numeric variables can be discretized. Missing values are discretized in bin number=max_bin+1. Default value is 25.
```
anneliesdej:
```suggestion
    :param max_bins: Maximum number categories in which numeric variables can be discretised. Missing values are discretised in bin number=max_bin+1. Default value is 25.
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
+import numpy as np
+import warnings
+from typing import Sequence
 
-def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_groups: int = 25, na_label: str = "__NA__") -> pd.DataFrame:
+__all__ = ["pairwise_spmse"]
+
+def _preprocessing_numeric(column: pd.Series, bins: Sequence[float] | None =None):
+    """
+    Preprocessing of the dataframes s.t. S_pMSE statistic can be calculated
```
anneliesdej: The current writing can be more formal/clear. Bins numeric in bins sounds a bit silly :). It could be something like
```
Preprocess a numeric column for S_pMSE computation by discretising it into bins.

Numeric values are assigned to bin intervals using the provided bin edges.
Missing values are preserved and are not modified.
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
+    Bins numerical values in bins
+    :param column: the specific column.
```
anneliesdej:
```suggestion
    :param column: Numeric column to be discretised.
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
 
-def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_groups: int = 25, na_label: str = "__NA__") -> pd.DataFrame:
+__all__ = ["pairwise_spmse"]
+
+def _preprocessing_numeric(column: pd.Series, bins: Sequence[float] | None =None):
+    """
+    Preprocessing of the dataframes s.t. S_pMSE statistic can be calculated
+    Bins numerical values in bins
+    :param column: the specific column. 
+    :param bins: array of bin edges
```
anneliesdej:
```suggestion
    :param bins: Bin edges used for discretisation. If `None`, no binning is applied.
    :return: Binned version of the input column.
```
Now it is also clear what happens when None is used.

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
+
+    if column.notna().any():
+        binned_column = pd.cut(column,bins)
+        column = binned_column
+    
+    return column
+
+def _preprocessing_non_numeric(column: pd.Series):
+    """
+    Preprocessing of the dataframes s.t. S_pMSE statistic can be calculated
```
anneliesdej:
```suggestion
    Standardise missing values in a non-numeric column.
    
    Missing entries are converted to np.nan to ensure consistent handling across categorical and string variables during S_pMSE computation.
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
+    return S_pMSE
+
+def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int = 25) -> pd.DataFrame:
     """
-    Compute the Standardized propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables 
+    Compute the Standardised propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables 
     between two similarly-structured dataframes: one original dataset and one synthetic version of the original dataset,
     assuming the same variables and the same number of rows.
-      
+
```
anneliesdej: I would probably rewrite the description ike this:
Compute the pairwise Standardised propensity Mean Squared Error (S_pMSE) between an original and a synthetic dataset.

The metric compares the preservation of pairwise joint distributions between corresponding variables in both datasets.

Numeric cariables are discretiesd into at most `max_bins` bins using bin edges derived jointly from the orignal and synthetic dataset. Categorical and string variables are used directly.

Missing values are treated as a separate category (np.nan).

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,177 @@
...
+        pd.DataFrame({"column1": ["c1"], "column2": ["c1"], "S_pMSE": [9/16]}), 25), 
+        #Check missing value handling, (str + pd.NA)
+
+        (pd.DataFrame({"c1": [np.nan, np.nan, np.nan, np.nan]}), 
+        pd.DataFrame({"c1": [np.nan, np.nan, np.nan, np.nan]}),
+        pd.DataFrame({"column1": ["c1"], "column2": ["c1"], "S_pMSE": [0.]}), 25),
+        #Full nan bin
+
+        (pd.DataFrame({"c1": [pd.NA, pd.NA, pd.NA, pd.NA]}), 
+        pd.DataFrame({"c1": [np.nan, np.nan, np.nan, np.nan, np.nan]}),
```
anneliesdej: I would change this line to only None's so you demonstrate that it also supports that input

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,153 @@
...
+    Calculate the joint frequency tables for variable pairs.
+    :param df: Dataset
+    :type: DataFrame
+    :param col1: column name 1
+    :type: str
+    :param col2: column name 2
+    :type: str
+    """
+    if col1 == col2:
+        jf = df.groupby(col1, dropna=False).size().rename(col1)
```
anneliesdej: Running the code gives this warning:
FutureWarning('The default of observed=False is deprecated and will be changed to True in a future version of pandas. Pass observed=False to retain current behavior or observed=True to adopt the future default and silence this warning.')].

It is likely from the groupby

anneliesdej (changes requested):

src/synthpop/utility_metrics/spmse.py
KoomenErina: The `:type:` directives should be removed consistently in this file. It might contradict the typehints.

src/synthpop/utility_metrics/spmse.py
```diff
@@ -96,15 +100,21 @@ def _calc_spmse(jf_or: pd.Series, jf_syn: pd.Series, n_o: int, n_s: int):
...
-    Compute the Standardised propensity Mean Squared Error (S_pMSE) as defined in [1] for all pairs of variables 
-    between two similarly-structured dataframes: one original dataset and one synthetic version of the original dataset,
-    assuming the same variables and the same number of rows.
+    Compute the pairwise Standardised propensity Mean Squared Error (S_pMSE) between an original and a synthetic dataset.
+
+    The metric compares the preservation of pairwise joint distributions between corresponding variables in both datasets.
+
+    Numeric variables are discretised into at most `max_bins` bins using bin edges derived jointly from the orignal and synthetic dataset. Categorical and string variables are used directly.
+
+    Missing values are treated as a separate category (np.nan).
```
KoomenErina: an example is missing

KoomenErina (changes requested):

KoomenErina (approved):

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
 This module contains metrics to evaluate the utility of synthetic data.
 """
 import pandas as pd
```
anneliesdej: Preferred import order is standard, third party, local. This would result in:
```python
from itertools import combinations_with_replacement
from typing import Sequence
import warnings

import numpy as np
import pandas as pd
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
+        * n_s / (n_o + n_s)
+
+    num_independent_combinations = len(expected_frequency.index)
+
+    if num_independent_combinations > 1:
+
+        S_pMSE = 1 / (num_independent_combinations - 1)\
+            * np.sum(
+            np.power(rescaled_differences, 2)
+            / expected_frequency
```
anneliesdej: The functional description mentions "Only category pairs with strictly positive expected frequency are retained for subsequent calculations.". So expected_frequency should be filtered on positives only. This way you also remove the risk of divison by zero in this line. This seems to be implicitly handled, but to match FD I would add it expliclitly

Edit after reviewing FD once more: k is defined twice in the FD, but both as the same. Remove the first definition, then the code is in line with FD

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
+
+    return S_pMSE
+
+
+def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int = 25) -> pd.DataFrame:
+    """
+    Compute the pairwise Standardised propensity Mean Squared Error (S_pMSE) between an original and a synthetic dataset.
+
+    The metric compares the preservation of pairwise joint distributions between corresponding variables in both datasets.
+    Numeric variables are discretised into at most `max_bins` bins using bin edges derived jointly from the orignal and synthetic dataset.
```
anneliesdej:
```suggestion
    Numeric variables are discretised into at most `max_bins` bins using bin edges derived jointly from the original and synthetic dataset. 
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
+        jf_or,
+        fill_value=0
+    )\
+        * n_s / (n_o + n_s)
+
+    num_independent_combinations = len(expected_frequency.index)
+
+    if num_independent_combinations > 1:
+
+        S_pMSE = 1 / (num_independent_combinations - 1)\
```
anneliesdej: The line continuation with a backslash is discouraged. Better would be:
```python
S_pMSE = (
    1 / (num_independent_combinations - 1)
    * np.sum(
        np.power(rescaled_differences, 2) / expected_frequency
    )
)
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
-    between two similarly-structured dataframes: one original dataset and one synthetic version of the original dataset,
-    assuming the same variables and the same number of rows.
-      
+
+    rescaled_differences = jf_syn.sub(
+        (n_s / n_o) * jf_or,
+        fill_value=0
+    )
+
+    expected_frequency = jf_syn.add(
```
anneliesdej: Here the backslash is used unnecessarily. You can change it to:
```python
expected_frequency = (
    jf_syn.add(jf_or, fill_value=0)
    * n_s / (n_o + n_s)
)
```

src/synthpop/utility_metrics/spmse.py
anneliesdej: PEP8 uses snake_case for variable names. S_pMSE does not follow this. It should be spmse or s_p_mse or s_pmse. Best would probably be spmse

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
+        S_pMSE = 1 / (num_independent_combinations - 1)\
+            * np.sum(
+            np.power(rescaled_differences, 2)
+            / expected_frequency
+        )
+
+    else:  # number_independent_combinations=1
+
+        S_pMSE = 0.
+        warnings.warn(
```
anneliesdej, line 98: PEP8 suggests 79ish characters on a line. (Guideline not a strict cut off). To adhere, you can cut the warning message like this:
```python
warnings.warn(
    f"Both variables are constant and equal; "
    f"the statistic is undefined. Returning 0 for variable pair: {jf_syn.name}"
)
```

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,289 @@
...
+        (
+            pd.DataFrame(
+                {
+                    "c1": [1, 0, np.nan], "c2": ['a', pd.NA, 'c'], "c3": [6, 7, 3]
+                }
+            ),
+            pd.DataFrame(
+                {
+                    "c2": [pd.NA, pd.NA, pd.NA], "c3": [
+                        6, 3, 6], "c1": [np.nan, np.nan, 0]
```
anneliesdej: This line is hard to read because it is broken up like this. Either keep it one line or push "c": [ to the next line

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,289 @@
...
+
+    ]
+)
+def test_pairwise_spmse_raises_wrong_inputs(orig_df, syn_df, max_bins, error):
+
+    with pytest.raises(ValueError, match=error):
+        pairwise_spmse(orig_df, syn_df, max_bins)
+
+
+@pytest.mark.parametrize(
```
anneliesdej, line 61: The big amount of different inputs make this test hard to read. You want a test that checks if the output is correct in format and number. You want a different test that checks if missing is handled correctly. Different column orders is a specific part of the FD, so I would make that a separate test with its own name. It is easier to read through the different tests if tests with different purposes have their own name. So this big test should be on if different types or sizes of inputs give the correct output. Then the other tests can be their own (parametrised) test.

This will also result in less tests that are dependent of hand-calculated results

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,289 @@
...
+        }
+    )
+
+    orig_before = orig.copy(deep=True)
+    syn_before = syn.copy(deep=True)
+
+    pairwise_spmse(orig, syn)
+
+    pd.testing.assert_frame_equal(orig, orig_before)
+    pd.testing.assert_frame_equal(syn, syn_before)
```
anneliesdej, line 367: Tests I would add:
- one that checks if the spmse of df1, df2 would be the same as df2, df1. This will guarantee the symmetry that is mentioned in the FD
- more as a sanity check one for scaling invariance: the S_pMSE should be invariant to uniform replication or scaling of either dataset. So duplicating rows in the original or synthetic data should not change any pairwise S_pMSE value and that for identical datasets it yields 0 for all variable pairs. (Basically spmse of small small should give the same result as spmse of small big where big is the same as small but with duplicated rows)
- a test with an extreme skew in the data
- a test that checks no-divison-by-zero explosion in rare cells.You can check the (awful quality) case of that skew being exact opposite in the synthetic data. Think 1000 zeros and 1 one vs 1 zero and 1000 ones
- a test that shows no NaN/inf explosions when a combination only appears in one dataset

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
+    """
+
+    if column.notna().any():
+        binned_column = pd.cut(column, bins)
+        column = binned_column
+
+    return column
+
+
+def _preprocessing_non_numeric(column: pd.Series):
```
anneliesdej:
```suggestion
def _preprocessing_non_numeric(column: pd.Series) -> pd.Series:
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
+from itertools import combinations_with_replacement
+import numpy as np
+import warnings
+from typing import Sequence
 
-def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_groups: int = 25, na_label: str = "__NA__") -> pd.DataFrame:
+__all__ = ["pairwise_spmse"]
+
+
+def _preprocessing_numeric(column: pd.Series, bins: Sequence[float] | None = None):
```
anneliesdej:
```suggestion
def _preprocessing_numeric(column: pd.Series, bins: Sequence[float] | None = None) -> pd.Series:
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
+    :param column: the specific column. 
+    """
+
+    mask = pd.isna(column)
+    column.loc[mask] = np.nan
+
+    return column
+
+
+def _joint_frequencies(df: pd.DataFrame, col1: str, col2: str):
```
anneliesdej:
```suggestion
def _joint_frequencies(df: pd.DataFrame, col1: str, col2: str) -> pd.Series:
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -2,22 +2,194 @@
...
+    if col1 == col2:
+        jf = df.groupby(col1, dropna=False, observed=True).size().rename(col1)
+    else:
+        jf = df.groupby([col1, col2], dropna=False,
+                        observed=True).size().rename(col1+','+col2)
+
+    return jf
+
+
+def _calc_spmse(jf_or: pd.Series, jf_syn: pd.Series, n_o: int, n_s: int):
```
anneliesdej:
```suggestion
def _calc_spmse(jf_or: pd.Series, jf_syn: pd.Series, n_o: int, n_s: int) -> float:
```

anneliesdej (changes requested):

tests/unit/test_S_pMSE.py
casvanerpduo: Regarding point 2: Due to the definition of the S_pMSE invariance is only valid if the S_pMSE is equal to 0.
The Rescaled difference (3.3 in FD) is invariant to duplication of rows. But the expected frequency is not. As such, the S_pMSE is dependent on the number of rows for non-zero ouputs.

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,194 @@
...
+
     :param orig_df: Original dataset.
-    :type orig_df: pd.DataFrame
     :param syn_df: Synthetic dataset. Should have the same columns as the original dataset, in the same order and the same number of rows.
-    :type syn_df: pd.DataFrame
-    :param max_groups: Maximum number categories in which numeric variables can be discretized. Default value is 25.
-    :type max_groups: int
-    :param na_label: String value to be given to missing values. Default is __NA__
-    :type na_label: str
+    :param max_bins: Maximum number categories in which numeric variables can be discretised. Missing values are discretised in bin number=max_bin+1. Default value is 25.
```
anneliesdej:
```suggestion
    :param max_bins: Maximum number of categories in which numeric variables can be discretised. Missing values are discretised in bin number=max_bin+1. Default value is 25.
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,194 @@
...
+    0     sex     sex  1.333333
+    1     sex  income  2.666667
+    2  income  income  1.333333
     """
-    return pd.DataFrame()
\ No newline at end of file
+
+    if not isinstance(orig_df, pd.DataFrame) or not isinstance(syn_df, pd.DataFrame):
+        raise ValueError(
+            "Original and synthetic dataframes should both be a pandas dataframe")
```
anneliesdej:
```suggestion
            "Original and synthetic dataframes should both be a pandas DataFrame.")
```
for codebase coherence

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,194 @@
...
+
+    n_o = len(orig_df.index)
+    n_s = len(syn_df.index)
+
+    if n_o == 0 or n_s == 0:
+        raise ValueError(
+            'Both the original and synthetic dataframe should consist out of non-zero rows')
```
anneliesdej:
```suggestion
            'Both the original and synthetic dataframe must be non-empty.')
```

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,434 @@
...
+        # Check column names not equal
+
+        (
+            pd.DataFrame({"B": [10], "A": [20]}), pd.DataFrame(
+                {"A": [10], "A": [20]}), 35, "must have the same shape and column names."
+        ),
+        # Check for multiple columns having the same name
+
+        (
+            [], [], 12, "both be a pandas dataframe"
```
anneliesdej:
```suggestion
            [], [], 12, "both be a pandas DataFrame"
```

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,434 @@
...
+
+        (
+            pd.DataFrame([0]), pd.DataFrame([0]), -
+            12, "with value of at least 1."
+        ),
+        # Check for negative bins
+
+        (
+            pd.DataFrame(), pd.DataFrame(), 35,
+            "dataframe should consist out of non-zero rows"
```
anneliesdej:
```suggestion
            "dataframe must be non-empty"
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,194 @@
...
+        jf_syn.add(jf_or, fill_value=0)
+        * n_s / (n_o + n_s)
+    )
+    num_independent_combinations = len(expected_frequency.index)
+
+    if num_independent_combinations > 1:
+
+        spmse = (
+            1 / (num_independent_combinations - 1)
+            * np.sum(
```
anneliesdej, line 91: you could add dtype=np.float32 here. Then the output typehint of the function can also be changed to np.float32. (in np.sum)

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,194 @@
...
+                np.power(rescaled_differences, 2) / expected_frequency
+            )
+        )
+    else:  # number_independent_combinations=1
+
+        spmse = 0.
+        warnings.warn(
+            f"Both variables are constant and equal; "
+            f"so the statistic is undefined. Return 0 for variable pair: {jf_syn.name}")
```
anneliesdej:
```suggestion
            f"the statistic is undefined. Return 0 for variable pair: {jf_syn.name}",
            UserWarning,
            )
```
This makes the warning type explicit, which makes testing and future maintenance clearer

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,194 @@
...
+
+
+def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int = 25) -> pd.DataFrame:
+    """
+    Compute the pairwise Standardised propensity Mean Squared Error (S_pMSE) between an original and a synthetic dataset.
+
+    The metric compares the preservation of pairwise joint distributions between corresponding variables in both datasets.
+    Numeric variables are discretised into at most `max_bins` bins using bin edges derived jointly from the original and synthetic dataset. 
+    Categorical and string variables are used directly.
+    Missing values are treated as a separate category (np.nan).
```
anneliesdej:
```suggestion
    Missing values are standardised to np.nan and included in the frequency calculations as a separate level.
```
Matches the implementation better

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,194 @@
...
+            "Original and synthetic dataframes should both be a pandas dataframe")
+
+    n_o = len(orig_df.index)
+    n_s = len(syn_df.index)
+
+    if n_o == 0 or n_s == 0:
+        raise ValueError(
+            'Both the original and synthetic dataframe should consist out of non-zero rows')
+
+    if set(orig_df.columns) != set(syn_df.columns):
```
anneliesdej, line 159:
```suggestion
    if sorted(orig_df.columns) != sorted(syn_df.columns):
```
This is a safer validation as duplicated column names could pass through the previous one:
Adding the following to your raises error test, does not actually raise an error:
```python
(
            pd.DataFrame([[10, 20]], columns=["A", "B"]), pd.DataFrame(
                [[10, 20, 30]], columns=["A", "A", "B"]), 35, "must have the same shape and column names."
        ),
```

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,194 @@
 """
 This module contains metrics to evaluate the utility of synthetic data.
 """
+from itertools import combinations_with_replacement
+from typing import Sequence
+import warnings
+
+import numpy as np
 import pandas as pd
+__all__ = ["pairwise_spmse"]
```
anneliesdej, line 11:
```suggestion

__all__ = ["pairwise_spmse"]
```
PEP8 wants one blank line here

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,434 @@
...
+    output1 = pairwise_spmse(orig_df1, syn_df1, max_bins=3)
+    output2 = pairwise_spmse(orig_df2, syn_df2, max_bins=3)
+
+    assert output1['S_pMSE'].iloc[0] == (output2['S_pMSE'].iloc[2])
+    assert output1['S_pMSE'].iloc[1] == (output2['S_pMSE'].iloc[1])
+    assert output1['S_pMSE'].iloc[2] == (output2['S_pMSE'].iloc[0])
+    # Tests symmetry s.t. spmse(X,Y) == spmse(Y,X)
+
+
+def test_pairwise_spmse_scaling_invariance():
```
anneliesdej:
```suggestion
def test_pairwise_spmse_scaling_invariance_on_identical_distributions():
```

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,434 @@
...
+            "S_pMSE": [4/3, 8/3, 4/3, 3., 20/9, 0.]
+        }
+    )
+
+    output = pairwise_spmse(orig_df, syn_df, max_bins=3)
+    pd.testing.assert_frame_equal(
+        output, expected, check_exact=False, rtol=1e-9)
+
+
+def test_pairwise_spmse_symmetry():
```
anneliesdej, line 401: This is now a test on column order independence. A better test would be:
```python
def test_spmse_symmetry():
    df1 = pd.DataFrame({"a": [1, 2, 3], "b": ["x", "y", "z"]})
    df2 = pd.DataFrame({"a": [1, 2, 4], "b": ["x", "x", "z"]})

    r1 = pairwise_spmse(df1, df2)
    r2 = pairwise_spmse(df2, df1)

    pd.testing.assert_frame_equal(
        r1.sort_values(["column1", "column2"]).reset_index(drop=True),
        r2.sort_values(["column1", "column2"]).reset_index(drop=True),
    )
```

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,434 @@
...
+
+    output1 = pairwise_spmse(orig_df1, syn_df1, max_bins=3)
+    output2 = pairwise_spmse(orig_df2, syn_df2, max_bins=3)
+    
+    pd.testing.assert_frame_equal(
+        output1, output2, check_exact=False, rtol=1e-9)
+    
+    assert output1.equals(expected)
+
+def test_pairwise_spmse_no_division_by_zero():
```
anneliesdej: 1 million strings seems a bit too heavy for a unit test, which we want to be lightweight. I would change it to 1000, which should still trigger the same behaviour

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,434 @@
...
+        {
+            "column1": ["c1",],
+            "column2": ["c1"],
+            "S_pMSE": [4*10**12/(10**6+2)]
+        }
+    )
+
+    output = pairwise_spmse(orig_df, syn_df, max_bins=3)
+
+    assert np.isfinite(output['S_pMSE']).all()
```
anneliesdej, line 443: This statement already checks the goal of the test. The check on the expected is not needed

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,434 @@
...
+        # Check for unequal number of columns
+
+        (
+            pd.DataFrame({"A": [10], "B": [20]}), pd.DataFrame(
+                {"A": [10], "C": [20]}), 25, "must have the same shape and column names."
+        ),
+        # Check column names not equal
+
+        (
+            pd.DataFrame({"B": [10], "A": [20]}), pd.DataFrame(
```
anneliesdej: This does not actually test for columns with duplicate column names:
>>> pd.DataFrame({"A": [10], "A": [20]})                                                                               
    A                                                                                                                  
0  20     

What would test it is:
```python
(
            pd.DataFrame([[10, 20]], columns=["A", "A"]), pd.DataFrame(
                [[10, 20]], columns=["A", "A"]), 35, "must have the same shape and column names."
        ),
```
Although a different error message might be needed here as they actually have the same shape and column names

tests/unit/test_S_pMSE.py
```diff
@@ -0,0 +1,434 @@
...
+                {
+                    "column1": ["c1"],
+                    "column2": ["c1"],
+                    "S_pMSE": [0.]
+                }
+            ),
+        ),
+    ]  # Full nan bin, checks pd.NA and None
+)
+def test_pairwise_spms_missing_value_handling(orig_df: pd.DataFrame, syn_df: pd.DataFrame, expected: pd.DataFrame):
```
anneliesdej:
```suggestion
def test_pairwise_spmse_missing_value_handling(orig_df: pd.DataFrame, syn_df: pd.DataFrame, expected: pd.DataFrame):
```

anneliesdej (changes requested):

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,201 @@
...
+    num_independent_combinations = len(expected_frequency.index)
+
+    if num_independent_combinations > 1:
+
+        spmse = (
+            1 / (num_independent_combinations - 1)
+            * np.sum(
+                np.power(rescaled_differences, 2) / expected_frequency
+            )
+        ).astype(np.float32)
```
anneliesdej, line 94: np.sum has a parameter called dtype. It would be better to set it there, because now it is first cast to np.float64 and then to np.float32. Changing it to the parameter avoids the cast to np.float64

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,201 @@
...
+
+    if n_o == 0 or n_s == 0:
+        raise ValueError(
+            'Both the original and synthetic dataframe must be non-empty.')
+
+    if not orig_df.columns.is_unique or not syn_df.columns.is_unique:
+        raise ValueError(
+            "Original and synthetic dataframes must have unique column names.")
+
+    if sorted(orig_df.columns) != sorted(syn_df.columns):
```
anneliesdej: I previously asked you to change it to the sorted version, however this introcues an ordering inconsistency problem later in `for column_name in orig_df` and `for col1, col2 in combinations_with_replacement`. They both rely on original column order and the validation does not check for column order anymore. So now the following will pass but produce results in different column orderings:
```python
orig_df.columns = ["A", "B"]
syn_df.columns = ["B", "A"]
```

It can easily be fixed by setting them early:
```python
cols = sorted(orig_df.columns)
orig_df = orig_df[cols]
syn_df = syn_df[cols]
```

I would then maybe also change the check back to set() as the docstrings mention that the column order must be the same. However, keeping it with the sorted version, the column order no longer needs to be the exact same.

Given that there is a test on different column orders, I would change the docstring and state that the column order does not matter, but the columns should exist.

src/synthpop/utility_metrics/spmse.py
```diff
@@ -1,23 +1,201 @@
...
+    """
+    Calculate the joint frequency tables for variable pairs.
+
+    :param df: Dataset
+    :param col1: column name 1
+    :param col2: column name 2
+    """
+
+    if col1 == col2:
+        jf = df.groupby(col1, dropna=False, observed=True).size().rename(col1)
```
anneliesdej, line 58: I think observed should be set back to False (by setting it explicitly, the warning is also gone). Setting it to true drops unused categories, which can make sure that missing categorical levels may disappear. So if a category in orig is not present in syn it gets dropped when it should just be seen as undefined or missing according to the FD

tests/unit/test_S_pMSE.py
```diff
@@ -159,6 +159,9 @@ def test_pairwise_spmse_raises_wrong_inputs(orig_df, syn_df, max_bins, error):
 def test_pairwise_spmse_input_shapes_and_types(orig_df: pd.DataFrame, syn_df: pd.DataFrame, expected: pd.DataFrame):
 
     output = pairwise_spmse(orig_df, syn_df)
+    expected['S_pMSE'] = expected['S_pMSE'].astype(
```
anneliesdej, line 175: This bakes the  implementation behaviour into the test instead of verifying it. A better approach would be something like
```python
assert output["S_pMSE"].dtype = np.float32
```

tests/unit/test_S_pMSE.py
```diff
@@ -369,30 +377,29 @@ def test_pairwise_spmse_extensive_output():
...
+    r1 = pairwise_spmse(df1, df2)
+    r2 = pairwise_spmse(df2, df1)
 
-    assert output1['S_pMSE'].iloc[0] == (output2['S_pMSE'].iloc[2])
-    assert output1['S_pMSE'].iloc[1] == (output2['S_pMSE'].iloc[1])
-    assert output1['S_pMSE'].iloc[2] == (output2['S_pMSE'].iloc[0])
+    pd.testing.assert_frame_equal(
+        r1.sort_values(["column1", "column2"]).reset_index(drop=True),
+        r2.sort_values(["column1", "column2"]).reset_index(drop=True),
+    )
```
anneliesdej: I found an even stronger version of the test that explicitly aligns pairs, so that coincidental matches would not pass anymore:
```python
merged = r1.merge(r2, on=["column1", "column2"], suffixes=("_12", "_21"))
assert np.allclose(merged["S_pMSE_12"], merged["S_pMSE_21"])
```

tests/unit/test_S_pMSE.py
anneliesdej, line 61: There is one more dtype I would add to tests and that is booleans as they should be treated as numericals

anneliesdej (changes requested):

src/synthpop/utility_metrics/spmse.py
casvanerpduo: Ordering the columns may reduce the interpretability of the S_pMSE output, as it will be differently structured than the original dataset.

With the check DataFrame.columns.is_unique we make sure the column names have a multiplicity of 1, and set() has an internal hash structure which means that iff. the orig and synthetic dataset have the same column names the sets will be exactly the same. As such, the S_pMSE can only be calculated in cases the column names are unique, and both the original and synthetic dataset share the same column names

KoomenErina (approved):

src/synthpop/utility_metrics/spmse.py
anneliesdej: After checking, .add() makes sure that there are no variables lost by setting it to True. Keeping it this way is more efficient

anneliesdej (approved):

Issue #106, opened 2026-05-18
implement handling random numbers


KoomenErina:
see #93 and #95 

This ticket is about the implementation of the design made in  #93 and #95

KoomenErina:
The requirements are changed:
When the user does not provide a seed, a sufficiently random seed should be generated.

PR #121, opened 2026-06-03
82 implement synthesiser

KoomenErina:
implemented synthesiser class.

Cannot be merged, since it has not been tested with `CartMethod`.

**point of attention:** is the behaviour for the first column as expected? The first column should be sampled. That is implemented by passing a constant feature to `CartMethod`. An integration test has been written to assert that this indeed effectively samples the first column. 

The specifications regarding randomness are not implemented, because of #106.

EDIT:
- Some basic integration tests have been done, but more is required. To keep this pull request small, that work is for an other issue (see #135)
- Issue #130 has been found during integration tests, but will not be fixed in this pull request.


src/synthpop/synthesiser.py
```diff
@@ -36,6 +57,29 @@ def fit(self, X: pd.DataFrame, y=None) -> Self:
 
         :return: Fitted synthesiser.
         """
+
+        if not isinstance(X,pd.DataFrame):
+            raise ValueError("only pandas dataframes are supported")
```
anneliesdej:
```suggestion
            raise ValueError(f"X must be a pandas DataFrame, got {type(X)} instead.")
```

src/synthpop/synthesiser.py
```diff
@@ -22,7 +23,27 @@ def __init__(self, random_seed: int,
...
                  ) -> None:
-        pass
+        
+        self.default_syn_method = default_syn_method
+        self.column_order = column_order
+        self.special_syn_method = special_syn_method
+
+    def _get_model(self,y):
+
+        if not (self.default_syn_method is None):
```
anneliesdej: Using is not None is not fully in line with PEP8. It would be better to do if is None and else

src/synthpop/synthesiser.py
```diff
@@ -22,7 +23,27 @@ def __init__(self, random_seed: int,
                  default_syn_method: BaseSynthMethod | None = None,
                  special_syn_method: dict[str, BaseSynthMethod] | None = None,
```
anneliesdej:
```suggestion
                 special_syn_method: Dict[str, BaseSynthMethod] | None = None,
```
Coherence

src/synthpop/synthesiser.py
```diff
@@ -3,6 +3,7 @@
 """
 from typing import Self
```
anneliesdej:
```suggestion
from typing import Self, Dict
```
Coherence

src/synthpop/synthesiser.py
```diff
@@ -36,6 +57,29 @@ def fit(self, X: pd.DataFrame, y=None) -> Self:
...
+            else:
+                predictors = X[self.column_order_[0:i]]
+
+            model = self._get_model(y)
+
+            self.models_[self.column_order_[i]] = model.fit(predictors,X[y])
+
         return self
 
     def generate(self, n: int | None = None, random_state: int = 42) -> pd.DataFrame:
```
anneliesdej: In the init the parameter is called random_seed, but here it is random_state. Both of them are also not used

tests/unit/test_synthesiser.py
```diff
@@ -0,0 +1,210 @@
...
+        self.fit_y = self.fit_y + [y]
+
+        return self
+    
+    def transform(self, X):
+        self.transform_X = self.transform_X + [copy.copy(X)]
+        return self.transform_result
+    
+    def get_feature_names_out(self, input_features=None):
+        raise Exception()
```
anneliesdej: I would add a comment here that this choice for raising is made because the function needs to exist because of the BaseSynthMethod, but that it should not be used and therefore if it is called it should fail. And add a test about it

src/synthpop/synthesiser.py
anneliesdej: first_column_method does not exist

src/synthpop/synthesiser.py
```diff
@@ -52,4 +96,16 @@ def generate(self, n: int | None = None, random_state: int = 42) -> pd.DataFrame
...
         :return: Synthetic dataset
         """
-        return pd.DataFrame()
+
+        result = pd.DataFrame()
+
+        for i,y in enumerate(self.column_order_):
+
+            if i == 0:
+                pred = pd.DataFrame({"init":[0]*self.n_samples_})
```
anneliesdej: `n` is ignored here. It should be something like
```python
n_rows = self.n_samples_ if n is None else n

pred = pd.DataFrame({"init": [0] * n_rows})
```

src/synthpop/synthesiser.py
```diff
@@ -36,6 +57,29 @@ def fit(self, X: pd.DataFrame, y=None) -> Self:
 
         :return: Fitted synthesiser.
         """
+
```
anneliesdej, line 112: Specification mentions that an exception should be raised if the original dataframe is empty

src/synthpop/synthesiser.py
```diff
@@ -52,4 +96,16 @@ def generate(self, n: int | None = None, random_state: int = 42) -> pd.DataFrame
...
+        result = pd.DataFrame()
+
+        for i,y in enumerate(self.column_order_):
+
+            if i == 0:
+                pred = pd.DataFrame({"init":[0]*self.n_samples_})
+            else:
+                pred = result
+
+            new_syn_column = self.models_[y].transform(X=pred)
```
anneliesdej, line 194: We also need to document in BaseSynthMethod specifically that it needs to allow a different number of rows to be returned, otherwise this line implicitly will assume that `len(output) == len(pred)`

src/synthpop/synthesiser.py
```diff
@@ -36,6 +57,29 @@ def fit(self, X: pd.DataFrame, y=None) -> Self:
 
         :return: Fitted synthesiser.
         """
+
+        if not isinstance(X,pd.DataFrame):
+            raise ValueError("only pandas dataframes are supported")
+        if self.column_order is None:
```
anneliesdej, line 120: column order now accepts anything so silently omitting or duplicating columns can happen. It would be better to add a validation like `set(column_order_) == set(X.columns)` and `len(column_order_) == len(X.columns)`

src/synthpop/synthesiser.py
```diff
@@ -36,6 +57,29 @@ def fit(self, X: pd.DataFrame, y=None) -> Self:
...
+            self.column_order_ = X.columns.to_list()
+        elif all([isinstance(item, int) for item in self.column_order]):
```
anneliesdej: Does this not create an unnecessary list?
Could it not be:
```python
all(instance(item, int) for item in self.column_order)
```

src/synthpop/synthesiser.py
```diff
@@ -52,4 +96,16 @@ def generate(self, n: int | None = None, random_state: int = 42) -> pd.DataFrame
         
         :return: Synthetic dataset
         """
-        return pd.DataFrame()
+
+        result = pd.DataFrame()
+
```
anneliesdej, line 179: There is no fit check

tests/unit/test_synthesiser.py
```diff
@@ -0,0 +1,210 @@
...
+    synth.models_["c"] = StubSynthMethod(transform_result=expected_result["c"])
+
+    result = synth.generate()
+    assert isinstance(result,pd.DataFrame)
+    assert expected_result.equals(result)
+
+    assert synth.models_["c"].transform_X[0].equals(expected_initial_data)
+    assert synth.models_["a"].transform_X[0].equals(expected_result[["c"]])
+    assert synth.models_["b"].transform_X[0].equals(expected_result[["c","a"]])
+
```
anneliesdej: I would add these tests:
- test if using n in generate works
- test what happens with an invalid column order (duplicate columns, columns that don't exist, not enough columns)
- test if you generate before fitting
- test that verifies there is no data leakage between stages
- repeated executions (which may not be fully possible since the random state manager is not implemented)

src/synthpop/synthesiser.py
```diff
@@ -22,7 +23,27 @@ def __init__(self, random_seed: int,
                  default_syn_method: BaseSynthMethod | None = None,
                  special_syn_method: dict[str, BaseSynthMethod] | None = None,
                  ) -> None:
-        pass
+        
+        self.default_syn_method = default_syn_method
+        self.column_order = column_order
+        self.special_syn_method = special_syn_method
+
+    def _get_model(self,y):
```
anneliesdej:
```suggestion
    def _get_model(self, y):
```
PEP8 wants a space after a comma. I added the comment 3 times, but it would be faster if you check all the cases

src/synthpop/synthesiser.py
```diff
@@ -36,6 +57,29 @@ def fit(self, X: pd.DataFrame, y=None) -> Self:
 
         :return: Fitted synthesiser.
         """
+
+        if not isinstance(X,pd.DataFrame):
```
anneliesdej:
```suggestion
        if not isinstance(X, pd.DataFrame):
```
PEP8 wants a space after a comma

src/synthpop/synthesiser.py
```diff
@@ -36,6 +57,29 @@ def fit(self, X: pd.DataFrame, y=None) -> Self:
...
+        if self.column_order is None:
+            self.column_order_ = X.columns.to_list()
+        elif all([isinstance(item, int) for item in self.column_order]):
+            self.column_order_ = X.columns[self.column_order].to_list()
+        else:
+            self.column_order_ = self.column_order
+
+        self.models_ = {}
+        self.n_samples_ = X.shape[0]
+        for i,y in enumerate(self.column_order_):
```
anneliesdej:
```suggestion
        for i, y in enumerate(self.column_order_):
```
PEP8

src/synthpop/synthesiser.py
anneliesdej, line 65: Eventually examples need to be added

src/synthpop/synthesiser.py
KoomenErina: Omitting columns from the column order would exclude columns from synthesis. I am not sure that is a problem.
Including columns in `column_order` that are not in the dataset would cause an out of bounds exception, which would be a reasonable error message.

however, duplicated entries in `column_order` are a problem. It would overwrite a fitted model, causing unexpected behaviour. A check will be added for that.

src/synthpop/synthesiser.py
KoomenErina: Not all synthesis methods support a different number of rows. `CopyMethod` does not.

anneliesdej: Aw right I forgot about that one

src/synthpop/synthesiser.py
anneliesdej: Agreed about omitting. If the error message from out of bounds is informative enough, we can use that one, but I would add a test that checks if that error is raised and match the message so if it changes we will see it and can be sure that the error is still reasonable

src/synthpop/synthesiser.py
KoomenErina: the name `random_seed` is used. This will be implemented with #106

tests/unit/test_synthesiser.py
KoomenErina: generate with different `n` is tested and implemented. Testing for invalid `column_order` too. Fit check have been added. The last two are not things to test in this pull request.

src/synthpop/synthesiser.py
```diff
@@ -1,8 +1,11 @@
 """
 module for generating synthetic data
 """
-from typing import Self
+from typing import Self, Dict
```
anneliesdej, line 4: PEP8 asks for specific ordering (standard, third party, local). So it would be:
```python
from typing import Self, Dict

import numpy as np
import pandas as pd
from sklearn import clone

from synthpop.methods.base_synth import BaseSynthMethod
from synthpop.methods.cart_synth import CartMethod
```

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
 
     :param random_seed: seed for randomness.
     :param column_order: list of variable names or list of indexes to define the order in which the columns will be synthesised. Default is the column order of the original dataset.
```
anneliesdej, line 20: Keep an eye on the line length for PEP8 (79 characters-ish)

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
...
+
+
+        if self.column_order is None:
+            self.column_order_ = X.columns.to_list()
+        elif all(isinstance(item, int) for item in self.column_order):
+
+            self._validate_column_order_unique(self.column_order)
+
+            array_columns = np.array(self.column_order)
+            out_of_bounds = array_columns >= len(X.columns)
```
anneliesdej: This is good, but negative indices also need to be considered. Either block these or explicitly allow the negative indexing and document it (I think blocking might make more sense)

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
...
+            if out_of_bounds.any():
+                raise ValueError(f"The following indices of Synthesiser.column_order are out of bounds: {array_columns[out_of_bounds]}")
+            self.column_order_ = X.columns[self.column_order].to_list()
+        else:
```
anneliesdej, line 139: I think we need to explicitly block mixed column orders (e.g. [0, "b"]). The current error is confusing:
ValueError: The following columns of Synthesiser.column_order are not in the dataframe: [0]

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
...
+        4           2.0             b   1000.299988
+        5           3.0             c      0.000000
+        6           1.0             a      0.200000
+        7           3.0             c      0.000000
+        8           1.0             a     -3.400000
+        9           1.0             a      0.200000
+
     """
     def __init__(self, random_seed: int,
                  column_order: list[str] | list[int] | None = None,
```
anneliesdej, line 68: I'm not sure if it is intuitive to add the column order in the init instead of passing it as a parameter in the fit. When I tried to test something I automatically add it as a fitting parameter, which seems more sklearn alike to me. The same might go for the other parameters as well

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
...
+            if len(columns_not_in_df)> 0:
+                raise ValueError(f"The following columns of Synthesiser.column_order are not in the dataframe: {sorted(columns_not_in_df)}")
+            self.column_order_ = self.column_order
+
+        self.models_ = {}
+        self.n_samples_ = X.shape[0]
+        for i, y in enumerate(self.column_order_):
+
+            if i == 0:
+                predictors = pd.DataFrame({"init":[0]*X.shape[0]})
```
anneliesdej:
```suggestion
                predictors = pd.DataFrame({"init": np.zeros(X.shape[0], dtype=int)})
```
Not wrong, but could be cleaner

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
...
+
+        if n is None:
+            n_syn_rows = self.n_samples_
+        else:
+            n_syn_rows = n
+
+        for i,y in enumerate(self.column_order_):
+
+            if i == 0:
+                pred = pd.DataFrame({"init":[0]*n_syn_rows})
```
anneliesdej:
```suggestion
                pred = pd.DataFrame({"init": np.zeros(n_syn_rows, dtype=int)})
```
Also see my earlier comment. Code is not wrong but this is cleaner

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
...
+        else:
+            model = effective_default_method
+
+        return model
+
+
+    def _validate_column_order_unique(self,column_order):
+        unique_column_order = np.unique_counts(column_order)
+
+        if (unique_column_order.counts >1).any():
```
anneliesdej:
```suggestion
        if (unique_column_order.counts > 1).any():
```
Keep an eye on the spacing for PEP8, this needs to be checked throughout the entire code. (Like line 92, 96, etc.)

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
...
-        using the previously generated columns as features.
-
         :param n: Number of rows to generate for the synthetic dataset. Default is the same number of rows than the dataset on which the synthesiser was fitted. If one of the synthesis methods copies the original data, this parameter must be None.
         :param random_state: Random seed generator. Default is 42. 
         
         :return: Synthetic dataset
         """
-        return pd.DataFrame()
+
+        if not hasattr(self,"models_"):
```
anneliesdej: If we want sklearn compatibility, check_is_fitted would be preferred

tests/unit/test_synthesiser.py
```diff
@@ -0,0 +1,275 @@
...
+    result = synth.generate()
+    assert isinstance(result,pd.DataFrame)
+    assert expected_result.equals(result)
+
+    assert synth.models_["a"].transform_X[0].equals(expected_initial_data)
+    assert synth.models_["b"].transform_X[0].equals(expected_result[["a"]])
+    assert synth.models_["c"].transform_X[0].equals(expected_result[["a","b"]])
+
+def test_generate_different_rowcount():
```
anneliesdej, line 247: This test verifies if the correct predictor size is passed into transform(), not if generate(n=5) returns a dataframe with 5 rows. The stub always returns 3. It reads a bit weird that the expected result has three rows when expected row count is set to 5

tests/integration/test_int_synthesiser.py
```diff
@@ -0,0 +1,174 @@
+import pytest
+import pandas as pd
+import numpy as np
+from synthpop.synthesiser import Synthesiser
+from synthpop.methods.cart_synth import CartMethod
+import string
+
```
anneliesdej, line 10: The integration tests are not yet tested with other synthesis methods (copy and sample)

tests/unit/test_synthesiser.py
```diff
@@ -0,0 +1,275 @@
...
+    not_a_df = {}
+    synth = Synthesiser(random_seed=3)
+
+    with pytest.raises(ValueError,match="X must be a pandas DataFrame, got <class 'dict'> instead."):
+        synth.fit(not_a_df)
+    
+def test_synthesiser_fit_throws_on_empty_dataframe():
+    df = pd.DataFrame()
+    synth = Synthesiser(random_seed=3)
+    with pytest.raises(ValueError,match="X can not be empty."):
```
anneliesdej:
```suggestion
    with pytest.raises(ValueError,match="X cannot be empty."):
```

src/synthpop/synthesiser.py
```diff
@@ -13,43 +16,169 @@ class Synthesiser:
...
         :param y: Ignored. This parameter exists only for compatibility with sklearn estimators.
 
         :return: Fitted synthesiser.
         """
+
+        if not isinstance(X, pd.DataFrame):
+            raise ValueError(f"X must be a pandas DataFrame, got {type(X)} instead.")
+        
+        if len(X) == 0:
+            raise ValueError("X can not be empty.")
```
anneliesdej:
```suggestion
            raise ValueError("X cannot be empty.")
```

anneliesdej (changes requested):

src/synthpop/synthesiser.py
KoomenErina: synthesiser is not sklearn compatible, see #13

tests/integration/test_int_synthesiser.py
KoomenErina: see #135

src/synthpop/synthesiser.py
KoomenErina: We look in to this during testing with users.

src/synthpop/synthesiser.py
KoomenErina: decided to let it be non-compatible with sklearn for now. That policy will be applied consistently to `Synthesiser`

src/synthpop/synthesiser.py
```diff
@@ -13,43 +18,178 @@ class Synthesiser:
...
+
+            self._validate_column_order_unique(self.column_order)
+
+            array_columns = np.array(self.column_order)
+            out_of_bounds = (array_columns >= len(X.columns))
+            if out_of_bounds.any():
+                raise ValueError(
+                    f"The following indices of Synthesiser.column_order are out of bounds: {array_columns[out_of_bounds]}")
+            
+            negative_indices = array_columns <0
```
anneliesdej:
```suggestion
            negative_indices = array_columns < 0
```

src/synthpop/synthesiser.py
```diff
@@ -13,43 +18,178 @@ class Synthesiser:
...
+            if negative_indices.any():
+                raise ValueError(f"negative indices not allowed.")
```
anneliesdej: It would be more in line with the other error messages if you state it as "The following indices of Synthesiser.column_order are negative: .... Negative indices are not allowed."

src/synthpop/synthesiser.py
```diff
@@ -13,43 +18,178 @@ class Synthesiser:
...
         Generate a synthetic dataset of ``n`` rows. 
 
-        This method loops through the columns of ``X``, following ``column_order``, and calls the :py:meth:`transform` function of the synthesis method classes from ``default_syn_method`` and ``special_syn_method``.
-        
-        For the first column, the method specified in ``first_column_method`` is used. For the next ones, we use their respective fitted functions. Each column is predicted
-        using the previously generated columns as features.
+        This method loops through the columns of ``X``, following ``column_order``, and calls the :py:meth:`transform` function of the synthesis method objects as used in `fit`.
 
         :param n: Number of rows to generate for the synthetic dataset. Default is the same number of rows than the dataset on which the synthesiser was fitted. If one of the synthesis methods copies the original data, this parameter must be None.
         :param random_state: Random seed generator. Default is 42.
```
anneliesdej:
```suggestion
        :param random_seed: Random seed generator. Default is 42. 
```
While it will be actually implemented later, it would be better to already correct this one small thing :)

tests/integration/test_int_synthesiser.py
```diff
@@ -0,0 +1,199 @@
+import pytest
```
anneliesdej: preferred import order would be:
```python
import string

import numpy as np
import pandas as pd
import pytest

from synthpop.methods.cart_synth import CartMethod
```

src/synthpop/synthesiser.py
```diff
@@ -13,43 +18,178 @@ class Synthesiser:
...
-                 special_syn_method: dict[str, BaseSynthMethod] | None = None,
+                 special_syn_method: Dict[str, BaseSynthMethod] | None = None,
                  ) -> None:
-        pass
+
+        self.default_syn_method = default_syn_method
+        self.column_order = column_order
+        self.special_syn_method = special_syn_method
+
+    def _get_model(self, y):
```
anneliesdej:
```suggestion
    def _get_model(self, column_name: str) -> BaseSynthMethod:
```
changing y to column_name might make it more intuitive

src/synthpop/synthesiser.py
```diff
@@ -13,43 +18,178 @@ class Synthesiser:
...
+        if self.special_syn_method is None:
+            model = effective_default_method
+        elif y in self.special_syn_method:
+            model = clone(self.special_syn_method[y])
+        else:
+            model = effective_default_method
+
+        return model
+
+    def _validate_column_order_unique(self, column_order):
```
anneliesdej:
```suggestion
    def _validate_column_order_unique(self, column_order: list[str] | list[int]):
```

src/synthpop/synthesiser.py
```diff
@@ -13,43 +18,178 @@ class Synthesiser:
...
         """
+
+        if not isinstance(X, pd.DataFrame):
+            raise ValueError(
+                f"X must be a pandas DataFrame, got {type(X)} instead.")
+
+        if len(X) == 0:
+            raise ValueError("X cannot be empty.")
+
+        if self.column_order is None:
```
anneliesdej, line 120: should there be anything for column_order=[]? It would be a silly to put it in as a parameter, so it could also be fine if we just let it return an empty dataframe as it does currently

src/synthpop/synthesiser.py
```diff
@@ -13,43 +18,178 @@ class Synthesiser:
...
+
+        if not hasattr(self, "models_"):
+            raise NotFittedError("synthesiser has not been fitted.")
+
+        result = pd.DataFrame()
+
+        if n is None:
+            n_syn_rows = self.n_samples_
+        else:
+            n_syn_rows = n
```
anneliesdej, line 185: I would add a ValueError if n is negative, otherwise np.zeros throws a NumPy exception

tests/unit/test_synthesiser.py
```diff
@@ -0,0 +1,295 @@
...
+    assert expected_result.equals(result)
+
+    assert synth.models_["c"].transform_X[0].equals(expected_initial_data)
+    assert synth.models_["a"].transform_X[0].equals(expected_result[["c"]])
+    assert synth.models_["b"].transform_X[0].equals(
+        expected_result[["c", "a"]])
+
+
+def test_generate_raises_when_not_fitted():
+    X = pd.DataFrame({"a": [0]})
```
anneliesdej:
```suggestion
```
Unused so can be removed

tests/unit/test_synthesiser.py
```diff
@@ -0,0 +1,295 @@
+import pytest
```
anneliesdej, line 6: Preferred import order would be
```python
import copy
import re

import pandas as pd
import pytest

from synthpop.methods.base_synth import BaseSynthMethod
```

tests/integration/test_int_synthesiser.py
```diff
@@ -0,0 +1,199 @@
...
+import pandas as pd
+import numpy as np
+from synthpop.synthesiser import Synthesiser
+from synthpop.methods.cart_synth import CartMethod
+import string
+
+
```
anneliesdej, line 11: I miss a test that verifies default behaviour in generate like:
```python
syn_df = synthesiser.fit(original_data).generate()
assert syn_df.shape[0] == original_data.shape[0]
```

There is also no check if all columns that get returned have the same length

tests/unit/test_synthesiser.py
```diff
@@ -0,0 +1,295 @@
...
+    assert synth.models_["a"].transform_X[0].equals(expected_result[["c"]])
+    assert synth.models_["b"].transform_X[0].equals(
+        expected_result[["c", "a"]])
+
+
+def test_generate_raises_when_not_fitted():
+    X = pd.DataFrame({"a": [0]})
+    synth = Synthesiser(random_seed=0)
+    with pytest.raises(NotFittedError):
+        synth.generate()
```
anneliesdej, line 314: I also would like to see a test if generate raises when invalid n is supplied. Also maybe a test on what happens (or is supposed to happen) when n is 0 or 1

tests/unit/test_synthesiser.py
```diff
@@ -0,0 +1,295 @@
...
+@pytest.mark.parametrize("column_order,expected_message",
+                         [
+                             (["a", "a", "b", "b", "c"],
+                              "The following columns occur multiple times in Synthesiser.column_order: ['a' 'b']"),
+                             ([2, 2, 1, 1, 0], "The following columns occur multiple times in Synthesiser.column_order: [1 2]"),
+                             (["a", "d", "c", "x"],
+                              "The following columns of Synthesiser.column_order are not in the dataframe: ['d', 'x']"),
+                             ([0, 3, 2, 4], "The following indices of Synthesiser.column_order are out of bounds: [3 4]"),
+                             ([0, "b", 1], "invalid column order: [0, 'b', 1]"),
+                             ([0,1, -1], "negative indices not allowed."),
```
anneliesdej:
```suggestion
                             ([0, 1, -1], "negative indices not allowed."),
```

tests/integration/test_int_synthesiser.py
```diff
@@ -0,0 +1,199 @@
...
+            dropna=False, normalize=True)
+        synthetic_dist = syn_df[cat_col].value_counts(
+            dropna=False, normalize=True)
+        max_diff = np.max(np.abs(original_dist - synthetic_dist))
+        assert max_diff < 0.02
+
+
+def test_synthesiser_preserves_num_num_relation():
+    n_samples_orig = 1000
+    original_data, index_num, index_cat = simulate_realistic_dataset_correlations(
```
anneliesdej:
```suggestion
    original_data, _, _ = simulate_realistic_dataset_correlations(
```
indices are not used in this test so they would not need to be saved in an object

tests/integration/test_int_synthesiser.py
```diff
@@ -0,0 +1,199 @@
...
+    syn_df = synthesiser.fit(original_data).generate()
+
+    obs_corr = original_data[["first", "second"]].corr()["second"]["first"]
+    syn_corr = syn_df[["first", "second"]].corr()["second"]["first"]
+    assert np.abs(obs_corr - syn_corr) < 0.01
+
+
+def test_synthesiser_preserves_cat_num_relation():
+    n_samples_orig = 1000
+    original_data, index_num, index_cat = simulate_realistic_dataset_correlations(
```
anneliesdej:
```suggestion
    original_data, _, _= simulate_realistic_dataset_correlations(
```

tests/integration/test_int_synthesiser.py
```diff
@@ -0,0 +1,199 @@
...
+    syn_df = synthesiser.fit(original_data).generate()
+
+    syn_means = syn_df.groupby("third")["fourth"].mean()
+    obs_means = original_data.groupby("third")["fourth"].mean()
+    assert np.max(np.abs(syn_means - obs_means))/max(obs_means) < 0.05
+
+
+def test_synthesiser_preserves_cat_cat_relation():
+    n_samples_orig = 10000
+    original_data, index_num, index_cat = simulate_realistic_dataset_correlations(
```
anneliesdej:
```suggestion
    original_data, _, _ = simulate_realistic_dataset_correlations(
```

anneliesdej (changes requested):

src/synthpop/synthesiser.py
KoomenErina: That is what I would expect it to do.

src/synthpop/synthesiser.py
```diff
@@ -13,43 +18,179 @@ class Synthesiser:
...
-        return pd.DataFrame()
+
+        if not hasattr(self, "models_"):
+            raise NotFittedError("synthesiser has not been fitted.")
+
+        result = pd.DataFrame()
+
+        if n is None:
+            n_syn_rows = self.n_samples_
+        elif n <0:
```
anneliesdej:
```suggestion
        elif n < 0:
```

anneliesdej (approved):

PR #133, opened 2026-06-08
84 implement plot univariate distributions

anneliesdej:


src/synthpop/plotting/plot.py
```diff
@@ -1,19 +1,367 @@
...
+    single HTML document to support scrolling and browser-based search.
+
+    If a saving location is provided, the generated HTML document is written
+    to `univariate_distribution_comparison.html` in the specified directory.
+    If interactive mode is enabled, the resulting HTML document is
+    automatically opened in the default web browser. When no saving location
+    is provided, a temporary HTML file is created and opened only when
+    interactive mode is enabled.
+
+    Notes
```
anneliesdej, line 293: Is it better to write the notes above or below the parameters?

KoomenErina: Personally I would prefer below

KoomenErina:
The current code is on the right track regarding outputting and showing.
However, as the code is currently, it is not possible to test the logic of how the data should be visualized (bins, dealing with missings, plot parameters,...). That is a problem because that is an important aspect of this feature. 

The reason it is not possible to test that is because this is one big function.

 It can be split up in the following responsibilities:
- making histograms
- making barplots
- combining barplots/histograms to a figure. (plotting 1 column)
- looping over the columns to make a figure
- outputting the figures to html (and showing the html if appropiate)

by splitting the large function up in smaller functions according to these tasks, each individual function becomes easily testable. 

See 84-suggestion-for-code-layout for how this can be done. (both code layout and tests)

KoomenErina:
Multiple tests aim to assert that a function is called. It is prefered to assert that the function has been called with the correct arguments. The call arguments are already stored

KoomenErina:
Multiple tests make assertions on the result html by testing for a substring. This is not an effective way of testing: "age" is a substring of "<page>". See the first comment for how to properly test this

KoomenErina:
Note that plotly seems to make it possible to define an entire figure from nested dictionaires. It migth be worthwhile to look in to that, since that might make testing even easier.

src/synthpop/plotting/plot.py
```diff
@@ -1,19 +1,367 @@
...
+        fig.update_layout(
+            legend_title_text="Dataset",
+            height=500,
+            margin=dict(b=100),
+        )
+
+        html_parts.append(
+            to_html(
+                fig,
+                include_plotlyjs="cdn",
```
KoomenErina: Loading plotly.js from their cdn implies that it only works with an internet connection. Furthermore, If i understand the docs correctly, the javascript is loaded for every figure.

I would recommend setting `include_plotlyjs` to `True` for the first figure, and then setting it to `False` for the others.

KoomenErina (changes requested):

KoomenErina:
I would like to suggest that `plot_univariate_distributions` returns the list of figures. That would probably work better with notebooks, and allows the user to do further things with the plots.

tests/unit/test_plot.py
```diff
@@ -0,0 +1,390 @@
...
+    def fake_bars(orig, syn):
+        called["bars"] += 1
+        return go.Bar(), go.Bar()
+
+    monkeypatch.setattr("synthpop.plotting.plot._make_histograms", fake_hist)
+    monkeypatch.setattr("synthpop.plotting.plot._make_bars", fake_bars)
+
+    fig = _plot_single_distribution(orig, syn, "age")
+
+    assert called["hist"] == 1
```
KoomenErina, line 164: test that `_make_histograms` has been called with the correct arguments.

tests/unit/test_plot.py
```diff
@@ -0,0 +1,390 @@
...
+    def fake_bars(orig, syn):
+        called["bars"] += 1
+        return go.Bar(), go.Bar()
+
+    monkeypatch.setattr("synthpop.plotting.plot._make_histograms", fake_hist)
+    monkeypatch.setattr("synthpop.plotting.plot._make_bars", fake_bars)
+
+    fig = _plot_single_distribution(orig, syn, "sex")
+
+    assert called["bars"] == 1
```
KoomenErina, line 192: test that `_make_bars` has been called with the correct arguments.

tests/unit/test_plot.py
```diff
@@ -0,0 +1,390 @@
...
+    fig = go.Figure()
+
+    def fake_to_html(fig, include_plotlyjs, full_html):
+        return "<div>FIG</div>"
+
+    monkeypatch.setattr("synthpop.plotting.plot.to_html", fake_to_html)
+
+    html = _build_html([fig])
+
+    assert "<div>FIG</div>" in html
```
KoomenErina: The  logic for `include_plotlyjs` is missing. The fact that `full_html` is `False` is should be tested.

tests/unit/test_plot.py
```diff
@@ -0,0 +1,390 @@
...
+def test_make_histograms_corrects_bins_for_integer_data():
+    orig = pd.Series([1, 2, 3])
+    syn = pd.Series([4, 5, 6])
+
+    orig_hist, syn_hist = _make_histograms(orig, syn)
+
+    # shared binning logic
+    assert orig_hist.xbins.start == 0.5
+    assert orig_hist.xbins.end == 6.5
+    assert orig_hist.xbins.size == 1
```
KoomenErina: `syn_hist` should be tested too

tests/unit/test_plot.py
```diff
@@ -0,0 +1,390 @@
...
+    _make_bars,
+    _plot_single_distribution,
+    _build_html,
+    _write_html,
+)
+
+
+# ----- _make_histograms tests -----
+
+def test_make_histograms_corrects_bins_for_integer_data():
```
KoomenErina, line 36: There are no test for non-integer numeric data. That has a different logic

tests/unit/test_plot.py
```diff
@@ -0,0 +1,390 @@
...
+
+    fig = _plot_single_distribution(orig, syn, "age")
+
+    annotation = fig.layout.annotations[0].text
+
+    assert annotation == (
+        "Missing values - Original: 1, Synthetic: 1"
+    )
+
+def test_plot_single_distribution_uses_column_name_in_title():
```
KoomenErina, line 211: There is more logic in the layout that needs to be tested:
- `barmode`
- axis titles

tests/unit/test_plot.py
```diff
@@ -0,0 +1,390 @@
...
+
+        def write(self, content):
+            self.content = content
+
+    monkeypatch.setattr("synthpop.plotting.plot.tempfile.NamedTemporaryFile", lambda *args, **kwargs: FakeTemp())
+
+    html = "<html></html>"
+    path = _write_html(html, None)
+
+    assert str(path).endswith(".html")
```
KoomenErina, line 316: assert that  ` NamedTemporaryFile` is used. This can be done by asserting against the `content` attribute of `FakeTemp`. It is also worthwhile to assert that `NameTemporaryFile` has been called with the right arguments. Things like utf8 and `delete` do not easily show up when visually inspecting the plots.

KoomenErina (changes requested):
The dependency on plotly should be in the pyproject.toml.

tests/unit/test_plot.py
```diff
@@ -19,19 +19,51 @@
...
 # ----- _make_histograms tests -----
 
+def test_make_histograms_returns_histograms():
+    orig = pd.Series([1, 2, 3])
+    syn = pd.Series([10, 11, 12])
+
+    orig_hist, syn_hist = _make_histograms(orig, syn)
+
+    assert isinstance(orig_hist, go.Histogram)
+    assert orig_hist.x.tolist() == orig.dropna().tolist()
```
KoomenErina, line 29: dropna is not needed, `orig` and `syn` do not have missing values

tests/unit/test_plot.py
```diff
@@ -19,19 +19,51 @@
...
-    assert orig_hist.xbins.start == 0.5
-    assert orig_hist.xbins.end == 6.5
-    assert orig_hist.xbins.size == 1
+    for histogram in [orig_hist, syn_hist]:
+        assert histogram.xbins.start == 0.5
+        assert histogram.xbins.end == 6.5
+        assert histogram.xbins.size == 1
+        assert histogram.histnorm == "probability density"
+
+def test_make_histograms_float_bin_logic():
```
KoomenErina: I see that `np.histogram_bin_edges` is doing the biggest part of the computations in this testcase.

I would recommend mocking `np.histogram_bin_edges` for this test, and setting it's return value.
This makes it feasable to assert for specific values of `xbins.size`/`xbins.start`/`xbins.end`.
If you do this, you also need to asset that  `np.histogram_bin_edges` has been called with the correct arguments.

tests/unit/test_plot.py
```diff
@@ -0,0 +1,494 @@
...
+    
+    assert len(mocked_environment["browser_calls"]) == 1
+    browser_args, _ = mocked_environment["browser_calls"][0]
+    assert browser_args[0] == (
+        Path("/fake/temp/univariate_distribution_comparison.html")
+        .resolve()
+        .as_uri()
+    )
+
+def test_plot_univariate_distributions_flow(monkeypatch):
```
KoomenErina, line 488: This is one of the test where my previous comment about asserting arguments applies. 
This fuction is a "moving data round" function. This test should assert that:
- the output of `_plot_single_distribution"` is in the input of `_build_html`
- the output of `_build_html` is in the input of `_write_html`
- the output of `_write_html` is in the input of `webbrowser.open`

src/synthpop/plotting/plot.py
```diff
@@ -1,27 +1,374 @@
 """
 This module contains functions to visually inspect synthetic data and evaluate its quality. 
 """
+from __future__ import annotations
```
KoomenErina: why is this needed?

KoomenErina (changes requested):
I tested this with the SD2011 and the resulting synthetic data.
Line 83 of plot.py resulted in the following error:
"TypeError: Cannot setitem on a Categorical with a new category (<MISSING>), set the categories first"

src/synthpop/plotting/plot.py
anneliesdej: Left from debugging an earlier version. No longer needed

KoomenErina (approved):

PR #136, opened 2026-06-09
83 implement plot spmse

casvanerpduo:
Request to implement the S_pMSE heatmap plotting.

Please take extra note to the hardcoded colormap/binning values. In the functional description these are pre-determined. But we could consider them to be a function argument.


src/synthpop/plotting/plot.py
```diff
@@ -2,6 +2,10 @@
 This module contains functions to visually inspect synthetic data and evaluate its quality. 
 """
 import pandas as pd
+import numpy as np
+import plotly.graph_objects as go
+import plotly.express as px
+import os
```
KoomenErina, line 12: import Figure

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,128 @@
...
+    fig = plot_spmse(spmse_df, None, False)
+    z = fig.data[0].z
+    assert np.shape(z) == (3, 3)
+
+def test_binning(spmse_df):
+    """Test that binned categories are correct (0-5)."""
+    fig = plot_spmse(spmse_df, None, False)
+    z = np.array(fig.data[0].z)
+    assert z.min() == 0
+    assert z.max() == 5
```
KoomenErina: test `fig.data[0].z` completely, not just the extreme values. Having less rows in the spmse dataframe makes this easier

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,128 @@
...
+
+def test_missing_text(spmse_df):
+    """Test that cells with 0 S_pMSE are shown as 'MISSING'."""
+    fig = plot_spmse(spmse_df, None, False)
+    text = np.array(fig.data[0].text)
+    assert "MISSING" in text
```
KoomenErina: test the full value of fig.data[0].text

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,128 @@
...
+    z = fig.data[0].z
+    assert np.shape(z) == (3, 3)
+
+def test_binning(spmse_df):
+    """Test that binned categories are correct (0-5)."""
+    fig = plot_spmse(spmse_df, None, False)
+    z = np.array(fig.data[0].z)
+    assert z.min() == 0
+    assert z.max() == 5
+
```
KoomenErina, line 245: test the values of fig.data[0] x and y

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,128 @@
...
+    colorbar = fig.data[0].colorbar
+    assert list(colorbar.ticktext) == [
+        "MISSING",
+        "(0,3]",
+        "(3,10]",
+        "(10,30]",
+        "(30,100]",
+        "(100,+)",
+    ]
+
```
KoomenErina, line 304: test for the correct color scales

KoomenErina (changes requested):

tests/unit/test_plot_spmse.py
```diff
@@ -123,6 +235,10 @@ def fake_show(*args, **kwargs):
     plot_spmse(spmse_df, None, False)
     assert not called
 
+
 def test_plot_visual(spmse_df):
```
KoomenErina: this test should be removed

src/synthpop/plotting/plot.py
```diff
@@ -19,66 +21,81 @@ def plot_univariate_distributions(obs_df: pd.DataFrame, syn_df: pd.DataFrame, ta
...
-        colorscale.append([(i+1)/n, color])
+
+        colorscale.append([i / n, color])
+        colorscale.append([(i + 1) / n, color])
 
     return colorscale
 
-def plot_spmse(spmse: pd.DataFrame, save_path: str | None, show_plot=True) -> Figure:
+
+def plot_spmse(spmse: pd.DataFrame, save_path: str | None, show_plot=True) -> go.Figure:
```
KoomenErina: add example

KoomenErina (changes requested):

src/synthpop/plotting/plot.py
```diff
@@ -68,6 +68,34 @@ def plot_spmse(spmse: pd.DataFrame, save_path: str | None, show_plot=True) -> go
     :param show_plot: Boolean index on whether the plot pops up in an interactive window
 
     :return: plotly.Figure
```
KoomenErina: You can describe the returned value. The type is already included in the documentation

src/synthpop/plotting/plot.py
KoomenErina: Mention here what happens if `save_path` is `None`

src/synthpop/plotting/plot.py
```diff
@@ -68,6 +68,34 @@ def plot_spmse(spmse: pd.DataFrame, save_path: str | None, show_plot=True) -> go
     :param show_plot: Boolean index on whether the plot pops up in an interactive window
```
KoomenErina: just a boolean, not an index

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def fake_show(*args, **kwargs):
+        nonlocal called
+        called = True
+
+    from plotly.graph_objects import Figure
+    monkeypatch.setattr(Figure, "show", fake_show)
+
+    plot_spmse(spmse_df, None, False)
+    assert not called
```
KoomenErina, line 417: The function changes the input, which is undesirable.
Add a test that asserts that the input arguments are not mutated

KoomenErina (changes requested):

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+
+    matrix = df.pivot(index="column1", columns="column2", values=value_string)
+    matrix = matrix.combine_first(matrix.T)
+
+    # invert s.t. the diagonal goes from upper-left to lower-right
+    return matrix.iloc[::-1]
+
+
+def _get_colorscale() -> list:
```
anneliesdej: British English would be colour scale

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+
+    for i, color in enumerate(colors):
+
+        colorscale.append([i / n, color])
+        colorscale.append([(i + 1) / n, color])
+
+    return colorscale
+
+
+def plot_spmse(spmse: pd.DataFrame, save_path: str | None = None, show_plot=True) -> go.Figure:
```
anneliesdej:
```suggestion
def plot_spmse(spmse: pd.DataFrame, save_path: str | None = None, show_plot: bool = True) -> go.Figure:
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Plot the standardised propensity mean squared error.
+
+    :param spmse: The standardised propensity mean squared error values. 
+    Should be a 3xN dataframe, where indices 0,1 are the column names and index 2 is the S_pMSE
+    :param save_path: File name and path to save the image of the plot. Does not save if None
+    :param show_plot: Boolean on whether the plot pops up in an interactive window
+
+    :return: A Plotly Figure containing a heatmap of pairwise S_pMSE values with
+    bin-based coloring and S_PMSE values in the bins.
```
anneliesdej:
```suggestion
    bin-based colouring and S_PMSE values in the bins.
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+
+    spmse = spmse.copy(deep=False)
+
+    bins = [0, 3, 10, 30, 100, np.inf]
+    bin_labels = [
+        "MISSING", "(0,3]", "(3,10]",
+        "(10,30]", "(30,100]", '(100,+)'
+    ]
+
+    spmse["category"] = np.digitize(spmse.iloc[:, 2], bins=bins, right=True)
```
anneliesdej: This implementation assumes pairwise_spmse does not return nan as a value, which is fine, but it is an assumption and therefore needs documentation

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+
+    matrix = _make_matrix(spmse, "category")
+    matrix_orig = _make_matrix(spmse, "S_pMSE")
+
+    # Preprocessing for Plotting
+    text_matrix = matrix_orig.round(2).astype(str)
```
anneliesdej: The FD writes that the diagonal should be empty

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+    text_matrix = text_matrix.mask(matrix_orig == 0, "MISSING")
+
+    colorscale = _get_colorscale()
+
+    # plotting
+    fig = go.Figure(
```
anneliesdej: I would consider writing the x axis above the plot. In my opinion, that would make it easier to read, or at least the diagonal

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+    ]
+
+    spmse["category"] = np.digitize(spmse.iloc[:, 2], bins=bins, right=True)
+
+    matrix = _make_matrix(spmse, "category")
+    matrix_orig = _make_matrix(spmse, "S_pMSE")
+
+    # Preprocessing for Plotting
+    text_matrix = matrix_orig.round(2).astype(str)
+    text_matrix = text_matrix.mask(matrix_orig == 0, "MISSING")
```
anneliesdej: Missing seems a bit misleading as it might indicate that there was an error in the calculation.

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+
+    colorscale = _get_colorscale()
+
+    # plotting
```
anneliesdej: The tooltips right now also show z, which is confusing. Is it possible to not have this shown in the rendered version?

src/synthpop/plotting/plot_spmse.py
anneliesdej: interactive would be coherent with plot_univariate_distributions, although interactive there specifically makes a HTML and here it depends on if you are running a python or jupyter notebook file what happens, so show_plot could be correct as well

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+            title="S_pMSE Heatmap",
+            title_x=0.5,
+            font=dict(family="Arial", size=15),
+            width=900,
+            height=845,
+            yaxis=dict(scaleanchor="x", scaleratio=1)  # Make it a Cube
+        )
+    )
+
+    if save_path:
```
anneliesdej, line 233: I would align this with the plot_univariate_distributions. Right now passing "heatmap.png" gives a FIleNotFound error.
plot_univariate_distributions uses:
```python
output = Path(save_path)
output.parent.mkdir(
    parents=True,
    exist_ok=True
)

output_file = (
        output_dir
        / "univariate_distribution_comparison.html"
    )
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+    >>> fig = plot_spmse(
+    ...     spmse=spmse,
+    ...     save_path=None,
+    ...     show_plot=True,
+    ... )
+    """
+
+    if not list(spmse.columns) == ['column1', 'column2', 'S_pMSE']:
+
+        raise ValueError(
```
anneliesdej, line 205: I would change the message to:
```python
raise ValueError(
    "The S_pMSE dataframe must contain exactly "
    "the columns ['column1', 'column2', 'S_pMSE']."
)
```

src/synthpop/plotting/plot_spmse.py
anneliesdej: Given that the column names are required, I would index on S_pMSE instead of :,2

src/synthpop/plotting/plot_spmse.py
anneliesdej: You can divide the code further in helpers like
- _categorise_spmse
- _make_matrix (already exists but adde for order)
- _make_text_matrix
- _get_colour_scale
- _make_heatmap

This way you can also monkey patch more

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,154 @@
...
+                outlinecolor="black",
+                outlinewidth=2,
+            )
+        ),
+
+        layout=dict(
+            title="S_pMSE Heatmap",
+            title_x=0.5,
+            font=dict(family="Arial", size=15),
+            width=900,
```
anneliesdej: These defaults are rather specific. It is not necessarily wrong but it stands out. Any reason why it should be these?

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,244 @@
...
+            )
+        ),  # Can only be exactly 3 columns
+
+        pd.DataFrame(
+            {
+                "column1": ["c1"],
+                "column2": ["c1"],
+                "spmse": [1.0]
+            }
+        )  # Wrong capitalization
```
anneliesdej:
```suggestion
        )  # Wrong capitalisation
```

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,244 @@
+import numpy as np
+import pytest
+import pandas as pd
+
+from synthpop.plotting.plot_spmse import plot_spmse
+
+
```
anneliesdej, line 144: I miss test that check if functions are called with the correct arguments. This is especially helpful given that plotly does most of the work

anneliesdej (changes requested):

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,244 @@
...
+    from plotly.graph_objects import Figure
+    monkeypatch.setattr(Figure, "write_image", fake_write_image)
+
+    outfile = tmp_path / "plot.png"
+    plot_spmse(spmse_df, str(outfile), False)
+
+    assert called
+
+
+def test_show_not_called(monkeypatch, spmse_df):
```
anneliesdej, line 402: You also want the opposite test that checks if it is called when True, with the correct arguments

src/synthpop/plotting/plot_spmse.py
casvanerpduo: After discussion with Betty the diagonal also contains usefull information. I will change the FD

src/synthpop/plotting/plot_spmse.py
casvanerpduo: Makes it a nice square without edges, height=width does not do so due the colorbar

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+    :param save_path: File directory to save the image of the plot. Does not save if None
+    :param show_plot: Boolean on whether the plot pops up in an interactive window
+
+    :return: A Plotly Figure containing a heatmap of pairwise S_pMSE values with
+    bin-based colouring and S_PMSE values in the bins.
+
+    Example:
+    --------
+    >>> import pandas as pd
+    >>> from synthpop.plotting.plot import plot_spmse
```
KoomenErina: import needs toe be updated

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+    :param matrix: NxN numpy matrix
+
+    return: NxN matrix
+    """
+    text_matrix = matrix.round(2).astype(str)
+    text_matrix = text_matrix.mask(matrix == 0, "CONSTANT VARIABLE")
+
+    return text_matrix
+
+def _get_colourscale() -> list:
```
anneliesdej: For proper British English I would change colourscale to colour_scale everywhere. So this function would be _get_colour_scale

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+
+    return fig
+
+def plot_spmse(spmse: pd.DataFrame, save_path: str | None = None, show_plot: bool = True) -> go.Figure:
+    """
+    Plot the standardised propensity mean squared error.
+
+    :param spmse: The standardised propensity mean squared error values.
+    Should be a 3xN dataframe, where indices 0,1 are the column names and index 2 is the S_pMSE
+    :param save_path: File directory to save the image of the plot. Does not save if None
```
anneliesdej:
```suggestion
    :param save_path: Directory where the output PDF is written.
        If a relative path is provided, it is resolved relative to the current working
        directory. The directory is created if it does not already exist (including)
        parent directories). If `None` (default), no permanent output file is created.
```
I would change it to this so it shows the same explanation as the plot_univariate_distributions

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+"""
+
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+import plotly.graph_objects as go
+import plotly.express as px
+
+def _categorise_spmse(spmse, bins):
```
anneliesdej: I would add typehints to the helper functions for coherence in the codebase. For example:
```python
def _categorise_spmse(
    spmse: pd.DataFrame,
    bins: Sequence[float],
) -> pd.DataFrame:
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+    Categorise the spmse.
```
anneliesdej:
```suggestion
    Categorise S_pMSE values into predefined bins.
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+def _categorise_spmse(spmse, bins):
+    """
+    Categorise the spmse. 
+    
+    This function assumes that the pairwise spmse does not return nan.
+
+    :param spmse: 3xN pandas DataFrame
+    :param bins: list of bin edges
+
+    :return: 4xN pandas DataFrame with new categorized column
```
anneliesdej:
```suggestion
    :return: 4xN pandas DataFrame with new categorised column
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+    
+    """
+    spmse["category"] = np.digitize(spmse['S_pMSE'], bins=bins, right=True)
+    return spmse
+
+def _make_matrix(df: pd.DataFrame, value_string="S_pMSE") -> pd.DataFrame:
+    """
+    Makes a Matrix of the 3xN S_pMSE array.
```
anneliesdej:
```suggestion
    Construct a symmetric matrix of the 3xN S_pMSE array.
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+    return: NxN matrix
+    """
+    text_matrix = matrix.round(2).astype(str)
+    text_matrix = text_matrix.mask(matrix == 0, "CONSTANT VARIABLE")
+
+    return text_matrix
+
+def _get_colourscale() -> list:
+    """"
+    Helper function to obtain the colourscale given the predetermined bins
```
anneliesdej:
```suggestion
    Helper function to obtain the discrete colour scale used for the S_pMSE bins.
```

Bit clearer

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+    4  Income     Sex    45.7
+    5     Sex     Sex     0.0
+    >>>
+    >>> fig = plot_spmse(
+    ...     spmse=spmse,
+    ...     save_path=None,
+    ...     show_plot=True,
+    ... )
+    """
+
```
anneliesdej, line 199: I would add a validation to check if it is a dataframe since python typehints will allow a different input. plot_univariate_distributions uses this validation:
```python
if not isinstance(orig_df, pd.DataFrame):
        raise ValueError(f"The original data should be a pandas DataFrame, got {type(orig_df)} instead.")
```
so you could probably add something like:
```python
if not isinstance(spmse, pd.DataFrame):
        raise ValueError(f"The S_pMSE data should be a pandas DataFrame, got {type(spmse)} instead.")
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+    bins = [0, 3, 10, 30, 100, np.inf]
+    bin_labels = [
+        "CONSTANT VARIABLE", "(0,3]", "(3,10]",
+        "(10,30]", "(30,100]", '(100,+)'
+    ]
+
+    # pairwise_spmse does not return nan
+    spmse = _categorise_spmse(spmse, bins)
+
+    #Make matrix for 2D visualisation
```
anneliesdej: This comment and the following two are probably not necessary. I would keep the "pairwise_spmse does not return nan" as it documents an assumption

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+            xaxis=dict(side="top"),
+            yaxis=dict(scaleanchor="x", scaleratio=1)  # Make it a Cube
+        )
+    )
+
+    return fig
+
+def plot_spmse(spmse: pd.DataFrame, save_path: str | None = None, show_plot: bool = True) -> go.Figure:
+    """
+    Plot the standardised propensity mean squared error.
```
anneliesdej: Since the docstring of plot_univariate_distributions is more detailed, I would expand this one as well to something like:
```
Create a heatmap visualisation of pairwise S_pMSE values.

The input dataframe is interpreted as a collection of pairwise relationships
between variables, where each row contains two variable names and their
associated S_pMSE value. A symmetric matrix is constructed from these
relationships and visualised as a heatmap, with variables shown on both axes.
The displayed S_pMSE values are grouped into predefined bins and
represented using a discrete sequential colour scale.

If a saving location is provided, the generated heatmap is written to `spmse.pdf` in the specified directory. If interactive rendering is enabled, the figure is displayed using the active Plotly renderer.
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+
+    :param spmse: The standardised propensity mean squared error values.
+    Should be a 3xN dataframe, where indices 0,1 are the column names and index 2 is the S_pMSE
+    :param save_path: File directory to save the image of the plot. Does not save if None
+    :param show_plot: Boolean on whether the plot pops up in an interactive window
```
anneliesdej:
```suggestion
    :param show_plot: Whether to display the heatmap interactively using the active Plotly renderer. Default is `True`. In headless environments this parameter should be set to `False`.
```
I did not check line length for this comment

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,228 @@
...
+        )
+    )
+
+    return fig
+
+def plot_spmse(spmse: pd.DataFrame, save_path: str | None = None, show_plot: bool = True) -> go.Figure:
+    """
+    Plot the standardised propensity mean squared error.
+
+    :param spmse: The standardised propensity mean squared error values.
```
anneliesdej:
```suggestion
    :param spmse: DataFrame containing the pairwise standardised propensity mean squared error values. The dataframe must contain exactly the columns `['column1', 'column2', 'S_pMSE']` where `column1` and `column2` identify the variable pair and `S_pMSE` contains the corresponding pairwise S_pMSE value. You can obtain this dataframe by running :func:`~synthpop.utility_metrics.spmse.pairwise_spmse`.
```
I did not check line length for this suggestion

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,438 @@
...
+
+    def fake_write_image(*args, **kwargs):
+        nonlocal called
+        called = True
+
+    from plotly.graph_objects import Figure
+    monkeypatch.setattr(Figure, "write_image", fake_write_image)
+
+    outfile = tmp_path / "spmse.pdf"
+    plot_spmse(spmse_df, str(outfile), False)
```
anneliesdej: The test is not wrong but it is mismatched with our parameters. Correct would be:
```python
plot_spmse(spmse_df, str(tmp_path), False)
```

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,438 @@
...
+    """
+
+    called = False
+
+    def fake_write_image(*args, **kwargs):
+        nonlocal called
+        called = True
+
+    from plotly.graph_objects import Figure
+    monkeypatch.setattr(Figure, "write_image", fake_write_image)
```
anneliesdej:
```suggestion
    monkeypatch.setattr(Figure, "plotly.graph_objects.Figure.write_image", fake_write_image)
```
safer patch

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,438 @@
...
+    """
+
+    called = False
+
+    def fake_show(*args, **kwargs):
+        nonlocal called
+        called = True
+
+    from plotly.graph_objects import Figure
+    monkeypatch.setattr(Figure, "show", fake_show)
```
anneliesdej, line 414:
```suggestion
    monkeypatch.setattr(Figure, "plotly.graph_objects.Figure.show", fake_show)
```
safer patch

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,438 @@
...
```
anneliesdej, line 432:
```suggestion
    monkeypatch.setattr(Figure, "plotly.graph_objects.Figure.show", fake_show)
```

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,438 @@
...
+                12.4598375983543,
+                4.0,
+                46.485343962786234,
+                0.0001,
+            ],
+        }
+    )
+
+
+# ----- _categorise_spmse test -----
```
anneliesdej, line 36: I would add one more tests that checks if the boundary conditions are correct. So you'd want to see the values 0, 3, 10 and 30

tests/unit/test_plot_spmse.py
```diff
@@ -0,0 +1,438 @@
...
+
+    fig = _make_heatmap(
+        matrix,
+        text_matrix,
+        colourscale,
+        bins,
+        bin_labels,
+    )
+
+    assert fig.data[0].hovertemplate == (
```
anneliesdej: Testing like this seems a little brittle to me. Formatting changes would break it. I think checking something like
```python
assert "%{x}" in fig.data[0].hovertemplate
assert "%{y}:" in fig.data[0].hovertemplate
assert "%{text}" in fig.data[0].hovertemplate
```
is also sufficient even though it checks less (as in you don't see if %{x} is not accidentally put behind y:)

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,250 @@
...
+        The dataframe must contain exactly the columns `['column1', 'column2', 'S_pMSE']` 
+        where `column1` and `column2` identify the variable pair and `S_pMSE` contains the corresponding pairwise S_pMSE value. 
+        You can obtain this dataframe by running :func:`~synthpop.utility_metrics.spmse.pairwise_spmse`.
+        Should be a 3xN dataframe, where indices 0,1 are the column names and index 2 is the S_pMSE
+    :param save_path: File directory to save the image of the plot. Does not save if None
+    :param show_plot: Whether to display the heatmap interactively using the active Plotly renderer. 
+        Default is `True`. In headless environments this parameter should be set to `False`.
+
+    :return: A Plotly Figure containing a heatmap of pairwise S_pMSE values with
+    bin-based colouring and S_PMSE values in the bins.
```
anneliesdej:
```suggestion
    bin-based colouring and S_pMSE values in the bins.
```

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,250 @@
+"""
+This module contains a function to visually inspect the quality of the multivariate SpMSE.
+"""
+
+from pathlib import Path
```
anneliesdej, line 5: from typing import Sequence needs to be added otherwise the tests cannot be found

src/synthpop/plotting/plot_spmse.py
```diff
@@ -0,0 +1,250 @@
...
+    represented using a discrete sequential colour scale.
+
+    If a saving location is provided, the generated heatmap is written to `spmse.pdf` in the specified directory. 
+    If interactive rendering is enabled, the figure is displayed using the active Plotly renderer.
+
+    :param spmse: DataFrame containing the pairwise standardised propensity mean squared error values. 
+        The dataframe must contain exactly the columns `['column1', 'column2', 'S_pMSE']` 
+        where `column1` and `column2` identify the variable pair and `S_pMSE` contains the corresponding pairwise S_pMSE value. 
+        You can obtain this dataframe by running :func:`~synthpop.utility_metrics.spmse.pairwise_spmse`.
+        Should be a 3xN dataframe, where indices 0,1 are the column names and index 2 is the S_pMSE
```
anneliesdej:
```suggestion
```
This line can be removed

Issue #130, opened 2026-06-05
bug: `MissingValuePredictor` seems to have a bias


KoomenErina:
When running `test_synthesiser_first_column_is_sampled_numeric`, the resulting proportions are consistently off by approx. 10%, this happens with 300 and 3000 samples.

**trigger**
Running `test_synthesiser_first_column_is_sampled_numeric` (test_int_synthesiser.py). This test tries to synthesise one numeric column, with the values 1.1, 2, `np.nan` in the proportions 1/2, 1/3, 1/6, and asserts that the synthetic column has the same proportions.

**Observation**
When inspecting what is happening in  `MissingValuePredictor.post_synth_transform`, the following observation can be made:

```
np.unique(y_val,return_counts=True,equal_nan=True)
>(array([1.1, 2. ], dtype=float32), array([1810, 1190]))

np.unique(y_out,return_counts=True,equal_nan=True)
>(array([1.10000002, 2.        ,        nan]), array([1810,  674,  516]))
```
**Note that the count of `1.1` did not change with setting `np.nan`**. Since half the rows are 1.1, this is highly unlikely. 
These variables (`y_out` and `y_val`) are observed just before exiting `MissingValuePredictor.post_synth_transform` (I put a breakpoint on line 249 of missing_value_handling.py)

**Expected behaviour**
The counts of 1.1 and 2 should both be reduced in equal percentages.

**good to know**

The related test `test_synthesiser_first_column_is_sampled_categorical` passes. The absolute difference in proportions is less than 5% in that test.

The root cause of this bug has not yet been found, it might be somewhere else than `MissingValuePredictor`

KoomenErina:
It appears that this problem is related to RNGs.

if you change line 75 of tree_utils.py from `self.random_state = random_state` to `self.random_state = np.random.default_rng()`, the problem disappears. 

Changing it to a non-reproducible RNG is not acceptable. This bug needs to be fixed with #106

KoomenErina:
In 130-bug-missingvaluepredictor-seems-to-have-a-bias, I added test aiming to reproduce the issue. With the same inputs as in the synthesiser integration test, the bug did not appear. Testing the sampler directly did not show this bug either.

PR #139, opened 2026-06-10
130 quickfix rng

KoomenErina:
a quick fix for #130 while #106 has not been completed yet.

The quick fix is to replace the default seed with a unseeded rng `np.random.default_rng()`. The consequence of this is that we lose reproducibility. That wil be restored in #106


anneliesdej (changes requested):
With this fix an integration test of CART no longer passes, while it should be passing (\tests\integration\test_int_cart_method.py::test_fit_transform_with_missing_values_in_target, y0)

1    <NA>
2       A
3       B
Name: target, dtype: string

    @pytest.mark.parametrize(
        "y",
        [
            pd.Series(["A", None, "A", "B"], dtype="string", name="target"),
            pd.Series([1, 2, np.nan, 4], name="target"),
            pd.Series([True, False, True, pd.NA], name="target"),
            pd.Series([1, None, 2, np.nan], name="target"),
        ],
    )
    def test_fit_transform_with_missing_values_in_target(y):
        X = pd.DataFrame(
            {"x1": [1, 2, 3, 4],
             "x2": ["a", "b", "y", "z"]}
        )
    
        cart = CartMethod()
        cart.fit(X, y)
    
        out = cart.transform(X)
    
        assert len(out) == len(y)
>       assert out.isna().sum() > 0
E       assert np.int64(0) > 0
E        +  where np.int64(0) = sum()
E        +    where sum = 0    False\n1    False\n2    False\n3    False\nName: target, dtype: bool.sum
E        +      where 0    False\n1    False\n2    False\n3    False\nName: target, dtype: bool = isna()
E        +        where isna = 0    B\n1    A\n2    B\n3    B\nName: target, dtype: object.isna

tests\integration\test_int_cart_method.py:264: AssertionError

tests/integration/test_int_cart_method.py
```diff
@@ -243,9 +243,9 @@ def test_fit_transform_with_missing_values_in_predictors():
 @pytest.mark.parametrize(
     "y",
     [
-        pd.Series(["A", None, "A", "B"], dtype="string", name="target"),
+        pd.Series(["A", None, None, "B"], dtype="string", name="target"),
```
anneliesdej: I would suggest leaving the probabilities as they are, but increase the number of rows so the test is less dependent on which rng is used

anneliesdej (commented):
There are now more tests that fail because they are dependent on which rng is used, like this one:
```
test_int_synthesiser.py::test_synthesiser_preserves_1D_statistics failed: def test_synthesiser_preserves_1D_statistics():
        """
        The aim of this test is to see if the means and univariate distributions are not wildly different.
        The goal is not to test that the synthetic data has a certain utility.
        The goal is to test that the synthetic data is reasonable. Benchmarking for utility should happen in other tests.
        """
        n_samples_orig = 1000
        n_samples_synthetic = 2000
        original_data, index_num, index_cat = simulate_realistic_dataset_correlations(
            n_samples=n_samples_orig)
        synthesiser = Synthesiser(random_seed=74124)
        syn_df = synthesiser.fit(original_data).generate(n=n_samples_synthetic)
        assert syn_df.shape[0] == n_samples_synthetic
        for num_col in index_num:
            original_mean = original_data[num_col].mean()
            synthetic_mean = syn_df[num_col].mean()
            assert np.abs(
                original_mean-synthetic_mean) > 1e-3, "original and synthetic are too close"
            assert np.abs(original_mean-synthetic_mean) / \
                original_mean < 0.05, "original and synthetic are too different"
        for cat_col in index_cat:
            original_dist = original_data[cat_col].value_counts(
                dropna=False, normalize=True)
            synthetic_dist = syn_df[cat_col].value_counts(
                dropna=False, normalize=True)
            max_diff = np.max(np.abs(original_dist - synthetic_dist))
>           assert max_diff < 0.02
E           assert np.float64(0.022499999999999964) < 0.02
 
tests\integration\test_int_synthesiser.py:159: AssertionError
```

It might be worth checking if this is an actual error or just an unlucky seed

tests/integration/test_int_synthesiser.py
```diff
@@ -53,7 +53,7 @@ def test_synthesiser_first_column_is_sampled_categorical():
         expected_proportions["missing"] - result_proportions[np.nan]) < 0.05
 
 
-@pytest.mark.xfail(reason="known issue, see #130")
+#@pytest.mark.xfail(reason="known issue, see #130")
```
anneliesdej, line 56: The synthesiser integration tests would benefit from an autopep8 run :)

anneliesdej (approved):

PR #149, opened 2026-06-16
146 implement randomstatemanager

KoomenErina:


src/synthpop/reproducibility.py
anneliesdej:
```suggestion
    The seed sequence is used to provide proper initialisation for all RNGs
    used in this package, even if the user provided seed is suboptimal.
```
Making it two lines is easier to read (and more PEP8 friendly)

anneliesdej:
```suggestion
        It is used to make instance seeds and seeds for
        external dependencies (legacy).
```

anneliesdej:
```suggestion
        The instance seeds are integers is to facilitate
        combining the root seed and instance seed.
```

src/synthpop/reproducibility.py
```diff
@@ -68,11 +75,13 @@ def create_new_seed(cls) -> int:
         ...     def fit(self,X,y):
         ...             self.random_state_ = RandomStateManager.create_new_seed()
         """
-        return cls._seed_sequence.spawn(1)[0].generate_state(1)
-
+        if cls._seed_sequence is None:
+            cls.set_root_seed(None)
+        # cls._seed_sequence.spawn(1)[0].generate_state(1)
```
anneliesdej:
```suggestion
```
This seems like old code that is no longer relevant and can therefore be removed

src/synthpop/reproducibility.py
```diff
@@ -81,18 +90,20 @@ def create_rng(cls,seed:int) ->np.random.Generator:
         The reason that the instance seeds are integers is to facilitate combining the root seed and instance seed.
         """
 
-        # default_rng uses seed sequences internally, so the easiest and safest way to combine is to pass a list of seeds.
-        return np.random.default_rng([cls._root_seed,seed])
+        if cls._root_seed is None:
+            cls.set_root_seed(seed=None)
 
+        # default_rng uses seed sequences internally, so the easiest and safest way to combine is to pass a list of seeds.
```
anneliesdej:
```suggestion
        # default_rng uses seed sequences internally
        # so the easiest and safest way to combine is to pass a list of seeds.
```

src/synthpop/reproducibility.py
```diff
@@ -81,18 +90,20 @@ def create_rng(cls,seed:int) ->np.random.Generator:
         The reason that the instance seeds are integers is to facilitate combining the root seed and instance seed.
         """
 
-        # default_rng uses seed sequences internally, so the easiest and safest way to combine is to pass a list of seeds.
-        return np.random.default_rng([cls._root_seed,seed])
+        if cls._root_seed is None:
```
anneliesdej, line 109: The functional description states that the root seed must be explicitly set before any RNG is created, otherwise an error should be raised. This is the opposite. However, if I remember correctly it was decided that not forcing setting a root seed was the way we would go, so the FD needs updating.

src/synthpop/reproducibility.py
```diff
@@ -81,18 +90,20 @@ def create_rng(cls,seed:int) ->np.random.Generator:
...
+            cls.set_root_seed(seed=None)
 
+        # default_rng uses seed sequences internally, so the easiest and safest way to combine is to pass a list of seeds.
+        return np.random.default_rng([cls._root_seed, seed])
```
anneliesdej, line 114: The FD states that "root seeds are processed through SeedSequence", which suggests the stored SeedSequence should participate in the RNG creation. However, the current method here is statistically fine so I would suggest updating the FD

src/synthpop/reproducibility.py
```diff
@@ -1,14 +1,18 @@
 import numpy as np
 from numpy.random import SeedSequence
+import secrets
 
 
+def _create_seed_from_sequence(seed_sequence) -> int:
```
anneliesdej: The return is actually a np.ndarray (np.uint32 specifically), not an int, so right now integer seeds are not actually integers which is not in line with the documentation and the type hints.

src/synthpop/reproducibility.py
anneliesdej, line 109: right now there is no defensive behaviour for invalid states, like passing "not an int", None or []. Invalid assumptions should fail safely (also the case in set_root_seed)

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,306 @@
+import pytest
+import numpy as np
+from synthpop.reproducibility import RandomStateManager
+
+import secrets
+
+"""
+The random state manager itself has 3 main states:
+1. uninitialized
```
anneliesdej: (un)initialize(d) must have the z changed to an s is all the files

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,306 @@
+import pytest
+import numpy as np
+from synthpop.reproducibility import RandomStateManager
+
+import secrets
+
```
anneliesdej: I would add tests that check:
- Context nesting (with RandomStateManager(1): with RandomStatemanager(2): ...)
- Re-entry (something like rsm.\_\_enter\_\_() rsm.\_\_enter()\_\_
- Seed = 0 and seed is very large (like 10 numbers)
- Test that checks if state after exit == state before enter

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,306 @@
...
+    def mock_set_seed(seed):
+        assert seed is None
+        monkeypatch.setattr(RandomStateManager, "_root_seed", expected_seed)
+        monkeypatch.setattr(RandomStateManager,
+                            "_seed_sequence", expected_seed_sequence)
+
+    monkeypatch.setattr(RandomStateManager, "set_root_seed", mock_set_seed)
+
+    expected_returned_seed = 44
+    mock_create_seed_from_sequence = mocker.patch(
```
anneliesdej, line 179: Is it not a bit too aggressive to mock internal seed generation?

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    first_seed1 = RandomStateManager.create_new_seed()
+    assert_create_new_seed_independence()
+    assert_rngs_reproducible()
+
+    # Assert that a random seed is used when no seed is given
+    rng1 = RandomStateManager.create_rng(3)  # RNG created with random seed
+
+    # Reset RandomStateManager
+    RandomStateManager._root_seed = None
```
anneliesdej: This seems like manually forcing internal state transitions instead of checking the system boundary. I would only modify the state through the public API (set_root_seed etc)

Adding RandomStateManager.reset() could also be helpful maybe

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,138 @@
+"""
+Integration tests of RandomStateManager.
+The aim is to test if RandomStateManager behaves as intended.
+
+Testing if the intended behaviour of RandomStateManager leads to reproducible synthetic data 
+is done in test_int_reproducibility.py
+"""
+import pytest
```
anneliesdej, line 9: I would add the following ones that clearly check some system contracts, independent of the synthesis. Some tests might already exist but they sometimes assert internal state instead of observable behaviour:
1. Same seed gives same output
```python
RandomStateManager.set_root_seed(42)

rng1 = RandomStateManager.create_rng(5)
rng2 = RandomStateManager.create_rng(5)

assert rng1.integers(100, size=10) == rng2.integers(100, size=10)
```

2. Different seeds give different streams
```python
rng1 = RandomStateManager.create_rng(1)
rng2 = RandomStateManager.create_rng(2)

assert rng1.integers(100, 10) != rng2.integers(100, 10)
```

3. Same root seed + same sequence gives sme derived seeds (if this is correct)
```python
RandomStateManager.set_root_seed(10)

a = RandomStateManager.create_new_seed()
b = RandomStateManager.create_new_seed()

RandomStateManager.set_root_seed(10)

a2 = RandomStateManager.create_new_seed()
b2 = RandomStateManager.create_new_seed()

assert a == a2
assert b == b2
```

4. Temporary override does not leak
```python
RandomStateManager.set_root_seed(1)

with RandomStateManager(999):
    rng_temp = RandomStateManager.create_rng(5)

rng_after = RandomStateManager.create_rng(5)

assert rng_temp.integers(100) != rng_after.integers(100)
```

These tests do not have _root_seed assertions or internal state inspection, but just check observable behaviour which I think is more fitting for integration tests

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def assert_create_new_seed_independence():
+    first_seed = RandomStateManager.create_new_seed()
+    second_seed = RandomStateManager.create_new_seed()
+
+    assert first_seed != second_seed, "create_new_seed does not produce independent seeds"
+
+
+def test_random_state_manager_default_random_seed():
```
anneliesdej: I might be incorrect, but this test seems like it is checking seed independence, RNG reproducibility, global reset behaviour and seed evolution, which seems like maybe too many responsibilities in one test

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,138 @@
...
+    RandomStateManager.set_root_seed(42)
+    sample4 = get_sample()
+    sample5 = get_sample()
+
+    assert sample1 == sample4, "RandomStateManager did not reset when setting the seed"
+    assert sample3 == sample5, "Exiting the context manager did not restore RandomStateManager"
+    assert sample2 != sample5, "Entering the context block did not change the seed"
+
+
+def test_random_state_manager_uninitialized_context_manager():
```
anneliesdej: This test fails for me:
E       AssertionError: exiting context block does not return RandomStateManager._root_seed to uninitialized
E       assert 42 is None
E        +  where 42 = RandomStateManager._root_seed

tests\integration\test_int_random_state_manager.py:124: AssertionError

tests/integration/test_int_reproducibility.py
```diff
@@ -0,0 +1,64 @@
+"""
+These test aim to show that the synthesis process is reproducible and deals correctly with randomness
```
anneliesdej, line 2: I expected some tests to use set_root_seed as well

src/synthpop/reproducibility.py
```diff
@@ -1,14 +1,18 @@
...
+    return seed_sequence.spawn(1)[0].generate_state(1)
+
 
 class RandomStateManager():
```
anneliesdej:
```suggestion
class RandomStateManager:
```
empty parantheses are unnecessary with PEP8

src/synthpop/reproducibility.py
```diff
@@ -68,11 +75,13 @@ def create_new_seed(cls) -> int:
         ...     def fit(self,X,y):
```
anneliesdej:
```suggestion
        ...     def fit(self, X, y):
```

src/synthpop/reproducibility.py
```diff
@@ -39,17 +43,20 @@ class RandomStateManager():
...
+
+        if seed is None:
+            cls._root_seed = secrets.randbits(128)
         else:
-            cls._seed_sequence = seed_sequence
+            cls._root_seed = seed
+
+        cls._seed_sequence = SeedSequence(cls._root_seed)
 
+        return
```
anneliesdej: This return can be removed

src/synthpop/reproducibility.py
anneliesdej:
```suggestion
def _create_seed_from_sequence(seed_sequence: SeedSequence) -> int:
```

src/synthpop/reproducibility.py
```diff
@@ -1,14 +1,18 @@
 import numpy as np
```
anneliesdej, line 5: Preferred import order is:
```python
import secrets

import numpy as np
from numpy.random import SeedSequence
```

tests/integration/test_int_reproducibility.py
```diff
@@ -0,0 +1,64 @@
...
+"""
```
anneliesdej, line 3:
```suggestion
"""

```
PEP8

tests/integration/test_int_reproducibility.py
```diff
@@ -0,0 +1,64 @@
...
+import pytest
```
anneliesdej, line 4: Preferred import order:
```python
import pytest

import numpy as np
from sklearn.base import BaseEstimator, TransformerMixin

from synthpop.reproducibility import RandomStateManager
```

tests/integration/test_int_reproducibility.py
```diff
@@ -0,0 +1,64 @@
...
+    This class implements randomness exactly as described in the developer docs.
+    It simulates how components in this package should use random numbers.
+    The aim is to test that the goals set out in the functional descriptions regarding randomness are met.
+    """
+
+    def __init__(self, random_state: int | None = None):
+        self.random_state = random_state
+
+    def fit(self, X, y):
+        self.random_state_ = RandomStateManager.create_new_seed(
```
anneliesdej: Preferred PEP8 formatting would be:
```python
if self.random_state is None:
    self.random_state_ = RandomStateManager.create_new_seed()
else:
    self.random_state_ = self.random_state
```

tests/integration/test_int_reproducibility.py
```diff
@@ -0,0 +1,64 @@
...
+        ) if self.random_state is None else self.random_state
+
```
anneliesdej, line 40: maybe add return self to follow sklearn contract?

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,138 @@
+"""
+Integration tests of RandomStateManager.
+The aim is to test if RandomStateManager behaves as intended.
+
+Testing if the intended behaviour of RandomStateManager leads to reproducible synthetic data 
+is done in test_int_reproducibility.py
+"""
```
anneliesdej, line 7:
```suggestion
"""

```
PEP8

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,306 @@
+import pytest
```
anneliesdej: I would write the doc string first, then a blank line, and then the import order:
```python
import secrets

import numpy as np
import pytest

from synthpop.reproducibility import RandomStateManager
```

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,138 @@
...
+
+    sample1 = rng1.integers(low=0, high=1000, size=100).tolist()
+    sample2 = rng2.integers(low=0, high=1000, size=100).tolist()
+
+    assert sample1 == sample2, "create_rng does not produce reproducible RNGs"
+
+    rng3 = RandomStateManager.create_rng(seed=2)
+    sample3 = rng3.integers(low=0, high=1000, size=100).tolist()
+
+    assert sample3 != sample1 and sample3 != sample2, "create_rng does not produce independent RNGs"
```
anneliesdej: Several places have some line length violations. You could write this ones as:
```python
assert sample3 != sample1 and sample3 != sample2, (
    "create_rng does not produce independent RNGs"
)
```

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,306 @@
...
+    """
+    Given: The random state manager is not initialized.
+    When: no root seed is provided (set_root_seed), seed is None, 
+    Then: The root seed is set with the value from secrets.randbits.
+    """
+
+    # Given: The random state manager is not initialized.
+    set_random_state_manager_state(None, None, monkeypatch)
+
+    # We patch secrets.randbits so that we can control the returned value and assert that it has been called correctly.
```
anneliesdej:
```suggestion
    # We patch secrets.randbits so that we can control the return value
    # and assert that it has been called correctly.
```

anneliesdej (changes requested):

tests/integration/test_int_random_state_manager.py
KoomenErina: This test fails inconsistently. The test passes if you only run this test, but fails if you run all integration tests at once. I suspect this has to do with a lack of cleanup/ test setup

src/synthpop/reproducibility.py
KoomenErina: Indeed, as mentionend in #106 , the requirement have changed. This will be updated in the FD.

src/synthpop/reproducibility.py
KoomenErina: The seed sequence is used to create instance seeds ( `create_new_seed`). `np.random.default_rng` uses seed sequences internally, so I guess the current implementation conforms to FD? 

But I agree, both are not clear. The FD will be updated, and more explanation will be added in the code.

tests/integration/test_int_random_state_manager.py
KoomenErina: There is no public api to reset the `RandomStateManager`.  A `reset` method is not something that would benefit the user, so it shouldn't be in the public api. 

To reliably test behaviour when `RandomStateManager` is not initialised, we need a way to reliably "uninitialize" it. This is probably also a problem in the failing integration test. 

So the options are:
- making an internal method to reset the state
- forcing the internal state as it is done now
- encapsulate checking if `RandomStateManager` is initialized in an internal method and mocking it in tests
- mocking _root_seed and _seed_sequence, which has the same problems as the current approach.

The current way does not require changes to the code of RandomState, and is effective and easy to understand. Mocking is an option, but would complicate these tests.

tests/unit/test_random_state_manager.py
KoomenErina: We need a way to assert that the seed sequence created by `set_root_seed` is used to generate the instance seed. The most direct method to test this requirement is to mock `RandomStateManager._seed_sequence`. 
But that would require mocking `SeedSequence.spawn`, and mocking the result of that function to assert the call to `generate_state`.  That would be a lot of test complexity to verify a statement with no variable parameters. 

That is why I made the `_create_seed_from_sequence` method, so that a simple mock would suffice in this test.

tests/integration/test_int_random_state_manager.py
KoomenErina: It was indeed the lack of cleanup/setup

tests/integration/test_int_random_state_manager.py
KoomenErina: We have decided to force the internal state in a test fixture. This requires no changes to `RandomStateManager` and can be done reliably.

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
...
         Set the root seed.
         The intended usage is within the Synthesiser class.
         """
-        cls._root_seed = seed
-        if seed_sequence is None:
-            cls._seed_sequence = SeedSequence(seed)
+
+        # The error messages of SeedSequence when seed is invalid are clear enough.
+
+        if len(seed) == 0:
```
anneliesdej:
```suggestion
        if isinstance(seed, (list, tuple, np.ndarray)) and len(seed) == 0:
```

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
...
+            warnings.warn("empty list as no entropy for seed. Use None for system entropy.",UserWarning)
```
anneliesdej:
```suggestion
            warnings.warn("empty list as no entropy for seed. Use None for system entropy.", UserWarning)
```

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
...
-        self.old_seed = None
 
     def __enter__(self):
         self.old_seed = RandomStateManager._root_seed
         self.old_seed_sequence = RandomStateManager._seed_sequence
         RandomStateManager.set_root_seed(self.new_seed)
 
-    def __exit__(self, type, value, traceback):
-        RandomStateManager.set_root_seed(self.old_seed ,self.old_seed_sequence)
+    def __exit__(self, type=None, value=None, traceback=None):
```
anneliesdej: Would it be better to use exc_type and exc_value. Because type is build in? It might not be an issue given that we don't use type() in this function

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
...
-        return np.random.default_rng([cls._root_seed,seed])
+        if cls._root_seed is None:
+            cls.set_root_seed(seed=None)
 
+        # default_rng uses seed sequences internally
+        # so the easiest and safest way to combine is to pass a list of seeds.
+        return np.random.default_rng([cls._root_seed, seed])
 
-    def __init__(self,seed):
+    def __init__(self, seed):
```
anneliesdej:
```suggestion
    def __init__(self, seed: int) -> None:
```

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
...
+        if seed is None:
+            cls._root_seed = secrets.randbits(128)
         else:
-            cls._seed_sequence = seed_sequence
+            cls._root_seed = seed
 
+        cls._seed_sequence = SeedSequence(cls._root_seed)
 
     @classmethod
     def create_new_seed(cls) -> int:
```
anneliesdej: This name might be slightly ambiguous as a user may ask new relative to what?

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
...
         """
         Returns a seed that can be used to make a RNG.
         The seed is based on the root seed.
-        It is used to make instance seeds and seeds for external dependencies (legacy).
-        The reason that the instance seeds are integers is to facilitate combining the root seed and instance seed.
+        It is used to make instance seeds and seeds for
+        external dependencies (legacy).
+        The instance seeds are integers is to facilitate
```
anneliesdej:
```suggestion
        The instance seeds are integers to facilitate
```

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
     """
     _seed_sequence = None
     """
-    The seed sequence is used to provide proper initialisation for all RNGs used in this package, even if the user provided seed is suboptimal.
+    The seed sequence is used to provide proper initialisation for all RNGs
```
anneliesdej: This docstring is not entirely true as creating a rng does not directly use _seed_sequence

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,333 @@
+"""
+The random state manager itself has 3 main states:
```
anneliesdej, line 2: I still miss a test on `with RandomStateManager`. I would even like to see one like:
```python
RandomStateManager.set_root_seed(1)

with RandomStateManager(2):
    with RandomStateManager(3):
        ...
```
Where you then check that inside inner is 3, after inner is 2 and after outer is 1

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,333 @@
...
+1. uninitialised
```
anneliesdej, line 3: There already are tests on enter and exit, but I think it is valuable to check if context is restored after an exception or error. Think something like:
```python
RandomStateManager.set_root_seed(1)

try:
    with RandomStateManager2):
        raise RuntimeError()
except RuntimeError:
    pass

assert root_seed == 1
```

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,333 @@
...
+
+def set_random_state_manager_state(root_seed, seed_sequence, monkeypatch):
+    """
+    Set the class variables of RandomStateManager,
+    """
+    monkeypatch.setattr(RandomStateManager, "_root_seed", root_seed)
+    monkeypatch.setattr(RandomStateManager, "_seed_sequence", seed_sequence)
+
+
+@pytest.fixture
```
anneliesdej, line 29: Given that default_rng is mocked, you want at least one test somewhere that checks if create_rng gives a np.random.Generator, just to validate assumptions. This can be in the integration tests as well as we test mocked assumptions there.

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,333 @@
...
+    # Then:
+    # - The root seed is set to the seed provided when creating an instance of RandomStateManager
+    assert RandomStateManager._root_seed == initial_seed, "initial seed is not restored"
+
+    # - The seed sequence is set to the same object as the old_seed_sequence property of the instance
+    assert RandomStateManager._seed_sequence is initial_seed_sequence, "initial seed is not restored"
+
+
+def test_random_state_manager_create_new_seed_returns_int():
+
```
anneliesdej, line 322: Given that tests should be independent, it would be more in line to reset the _rot_seed and _seed_sequence here as well

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,147 @@
...
+
+def test_random_state_manager_different_instance_seed_different_output():
+    RandomStateManager.set_root_seed(10)
+    rng1 = RandomStateManager.create_rng(1)
+    rng2 = RandomStateManager.create_rng(2)
+
+    assert get_sample_from_rng(rng1) != get_sample_from_rng(rng2)
+
+
+def test_random_state_manager_setting_root_seed_reproduces():
```
anneliesdej, line 76: It would be a stronger version (or separate test) to check if RNG regeneration after root seed reset gives the same samples as well, like
```
set_root_seed(10)
rng_a = create_rng(5)
set_root_seed(10)
rng_b = create_rng(5)
```
This test would connect the seed reproducibility to the downstream RNG reproducibility

anneliesdej (changes requested):

tests/unit/test_random_state_manager.py
KoomenErina: that test is in the integration tests. 

A context manager has 2 units: __enter__ and __exit__, both are tested here. How they work together is tested in the integration tests.

tests/unit/test_random_state_manager.py
KoomenErina: incorporated in an integration test

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
...
-        if seed_sequence is None:
-            cls._seed_sequence = SeedSequence(seed)
+
+        # The error messages of SeedSequence when seed is invalid are clear enough.
+
+        if isinstance(seed, (list, tuple, np.ndarray)) and len(seed) == 0:
+            warnings.warn("empty list as no entropy for seed. Use None for system entropy.", UserWarning)
+
+        if seed is None:
+            cls._root_seed = secrets.randbits(128)
```
casvanerpduo, line 70: Just to check: Does randbits(128) provide a 'good enough' random number?

In the package we store everything as float32, should we consider similar 32-bits datatypes here?

src/synthpop/reproducibility.py
```diff
@@ -35,64 +41,78 @@ class RandomStateManager():
...
         else:
-            cls._seed_sequence = seed_sequence
+            cls._root_seed = seed
 
+        cls._seed_sequence = SeedSequence(cls._root_seed)
 
     @classmethod
     def create_new_seed(cls) -> int:
         """
         Returns a seed that can be used to make a RNG.
```
casvanerpduo: I think correct spelling is an RNG

src/synthpop/reproducibility.py
```diff
@@ -1,14 +1,20 @@
+import secrets
+import warnings
+
 import numpy as np
 from numpy.random import SeedSequence
 
 
+def _create_seed_from_sequence(seed_sequence: SeedSequence) -> int:
+    return int(seed_sequence.spawn(1)[0].generate_state(1)[0])
```
casvanerpduo: Can this line be rewritten? I find it very complex

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+    result = RandomStateManager.create_new_seed()
+    assert isinstance(result, int)
+
+
+def test_random_state_manager_raises_on_invalid_type():
```
casvanerpduo, line 327: What happens when you pass a list, numpy array, dataframe (with mutiple datatypes)?

casvanerpduo: And if we accept lists/arrays, will the code be consistent and behave similair for all types?

src/synthpop/reproducibility.py
KoomenErina: Numpy recommends using `secrets.random`, see https://numpy.org/doc/stable/reference/random/index.html#recommend-secrets-randbits

In this package, we store the numeric float data that is observed/synthesised as float32. The reason for that is because the decision trees of sklearn preform better with float32. Using less bits here could have implications for "how random" the PRNG is.

tests/unit/test_random_state_manager.py
KoomenErina: lists and arrays are accepted. Numpy raises a instructive error message when an invalid argument is passed.

src/synthpop/reproducibility.py
KoomenErina: What would be a better name?

anneliesdej: Considerations could be create_instance_seed or spawn_seed I think

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
...
+            cls.set_root_seed(None)
+        return _create_seed_from_sequence(cls._seed_sequence)
 
     @classmethod
-    def create_rng(cls,seed:int) ->np.random.Generator:
+    def create_rng(cls, seed: int) -> np.random.Generator:
         """
-        Creates an RNG.
+        Creates a new instance of an RNG with a fixed initial state.
         Same root seed + same seed => same RNG.
```
anneliesdej:
```suggestion
        Same root seed + same seed => RNGs with identical random streams.
```
Little tweak to match reality better as create_rng(3) is not create_rng(3) if you were to test it

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
...
+        # default_rng uses seed sequences internally
+        # so the easiest and safest way to combine is to pass a list of seeds.
+        return np.random.default_rng([cls._root_seed, seed])
 
-    def __init__(self,seed):
+    def __init__(self, seed: int) -> None:
         self.new_seed = seed
-        self.old_seed = None
 
     def __enter__(self):
```
anneliesdej, line 119: Should this return self? That seems to be conventional: https://docs.python.org/3/library/contextlib.html. It returns the resource that needs to be managed (saw this on Medium and stackoverflow)

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
...
         self.old_seed = RandomStateManager._root_seed
         self.old_seed_sequence = RandomStateManager._seed_sequence
         RandomStateManager.set_root_seed(self.new_seed)
 
-    def __exit__(self, type, value, traceback):
-        RandomStateManager.set_root_seed(self.old_seed ,self.old_seed_sequence)
+    def __exit__(self, exc_type=None, exc_value=None, traceback=None):
```
anneliesdej:
```suggestion
    def __exit__(self, exc_type=None, exc_value=None, exc_traceback=None):
```
Just to be sure. This also seems conventional: https://www.geeksforgeeks.org/python/__exit__-in-python/

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
     """
     _seed_sequence = None
     """
-    The seed sequence is used to provide proper initialisation for all RNGs used in this package, even if the user provided seed is suboptimal.
+    The seed sequence is used to provide proper seeds for the classes that need an RNG
```
anneliesdej:
```suggestion
    The seed sequence is used to derive independent integer seeds for components that require their own RNG.
```
Seems a little more in line with the implementation

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,161 @@
+"""
```
anneliesdej, line 1: In one of the tests, maybe also a unit test, I would use root seed of 0, as this is a common seed people use

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+        - The root seed is set to the seed provided when creating an instance of RandomStateManager
+    """
+
+    # Given: The random state manager is not initialised.
+    set_random_state_manager_state(None, None, monkeypatch)
+
+    # When:
+    # an instance of RandomStateManager is created
+    instance = RandomStateManager(seed=5)
+    instance.__enter__()  # and __enter__ is called
```
anneliesdej, line 237: Is \_\_enter\_\_ the public API or with RandomStateManager? If the second is public API, I would use that in the test

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,161 @@
...
+Integration tests of RandomStateManager.
```
anneliesdej, line 2: I would add one more integration test:
```python
def test_random_state_manager_different_root_seed_different_output():

    RandomStateManager.set_root_seed(1)
    a = RandomStateManager.create_rng(5).integers(0,100,100)

    RandomStateManager.set_root_seed(2)
    b = RandomStateManager.create_rng(5).integers(0,100,100)

    assert not np.array_equal(a,b)
```
This test checks the following property of the functional description:
`Root-seed isolation: changing the root seed results in entirely different random streams`

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+    with pytest.raises(TypeError,match = ".* expects int or sequence of ints .*"):
+        RandomStateManager.set_root_seed(np.nan)
+
+def test_random_state_manager_warns_on_empty_list():
+
+    with pytest.warns(UserWarning) as record:
+        RandomStateManager.set_root_seed([])
+
+    assert len(record) == 1
+    assert str(record[0].message) =="empty list as no entropy for seed. Use None for system entropy."
```
anneliesdej:
```suggestion
    assert str(record[0].message) == "empty list as no entropy for seed. Use None for system entropy."
```
Eenie meenie style change

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+
+def test_random_state_manager_create_new_seed_returns_int():
+
+    result = RandomStateManager.create_new_seed()
+    assert isinstance(result, int)
+
+
+def test_random_state_manager_raises_on_invalid_type():
+
+    with pytest.raises(TypeError,match = ".* expects int or sequence of ints .*"):
```
anneliesdej:
```suggestion
    with pytest.raises(TypeError, match = ".* expects int or sequence of ints .*"):
```

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+        RandomStateManager.set_root_seed("not an integer")
+
+    with pytest.raises(TypeError,match = ".* expects int or sequence of ints .*"):
```
anneliesdej:
```suggestion
    with pytest.raises(TypeError, match = ".* expects int or sequence of ints .*"):
```

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+        RandomStateManager.set_root_seed(1.2)
+
+    with pytest.raises(ValueError,match = re.escape("expected non-negative integer")):
+        RandomStateManager.set_root_seed(-1)
+
+    with pytest.raises(TypeError,match = ".* expects int or sequence of ints .*"):
```
anneliesdej:
```suggestion
    with pytest.raises(TypeError, match = ".* expects int or sequence of ints .*"):
```

anneliesdej (changes requested):

src/synthpop/reproducibility.py
KoomenErina: In this situation, there is no resource that needs to be managed. Returning `self` would return an object with no methods or properties that the user should use. The current behaviour if that `with RandomStateManager(1111) as rsm` assigns `None` to rsm. That does not give a wrong signal.

Note that this situation is different from `fit`. Calls to `fit` are usually chained with other methods. Such chaining is only possible if fit returns `self`. Chaining is not possible with context managers.

tests/integration/test_int_random_state_manager.py
KoomenErina: added in some integration test

tests/unit/test_random_state_manager.py
KoomenErina: I would say it is a bit of grey area. Each of those special methods has a single responsibility, and are therefore tested individually. The user never calls `__enter__` and `__exit__` directly. But the `with ...` syntax is equivalent to calls to those functions. For this scenario, we treat `__enter__` and `__exit__` as public api

src/synthpop/reproducibility.py
```diff
@@ -1,14 +1,26 @@
...
+import warnings
+
 import numpy as np
 from numpy.random import SeedSequence
 
 
+def _create_seed_from_sequence(seed_sequence: SeedSequence) -> int:
+
+    # SeedSequence.generate_state returns the same value if you call it multiple times.
+    # To guarantee independent RNGS, we need independent seeds.
```
anneliesdej:
```suggestion
    # To guarantee independent RNGs, we need independent seeds.
```

src/synthpop/reproducibility.py
```diff
@@ -1,14 +1,26 @@
...
+    # That is why we need to generate a new seed sequence for each seed.
+    new_seed_sequence = seed_sequence.spawn(1)[0]
+    new_seed = new_seed_sequence.generate_state(1)[0] # generate_state generates a list of seeds, we need only one
```
anneliesdej:
```suggestion
    new_seed = new_seed_sequence.generate_state(1)[0]  # generate_state generates a list of seeds, we need only one
```
PEP8 spacing

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
     """
     _seed_sequence = None
     """
-    The seed sequence is used to provide proper initialisation for all RNGs used in this package, even if the user provided seed is suboptimal.
+    The seed sequence is used to derive independent integer seeds for components that require their own RNG.
```
anneliesdej: This whole docstring block reads awkwardly because the sentence is broken. I would change it to:
```
The seed sequence is used to derive independent integer seeds for
components that require their own RNG, even if the user provided
seed has poor statistical properties.
```

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
...
+            cls._root_seed = seed
 
+        cls._seed_sequence = SeedSequence(cls._root_seed)
 
     @classmethod
-    def create_new_seed(cls) -> int:
+    def create_instance_seed(cls) -> int:
         """
-        Returns a seed that can be used to make a RNG.
+        Returns a seed that can be used to make an RNG.
```
anneliesdej:
```suggestion
        Returns an independent integer seed derived from the root seed.
```
This is more in line with what this function actually does.

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
...
         The seed is based on the root seed.
```
anneliesdej:
```suggestion
```
If you apply the suggestion above, this sentence can be removed.

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
...
-        It is used to make instance seeds and seeds for external dependencies (legacy).
-        The reason that the instance seeds are integers is to facilitate combining the root seed and instance seed.
+        It is used to make instance seeds and seeds for
```
anneliesdej:
```suggestion
        It can also be used to initialise external dependencies that require integer seeds.
```
The instance seed part is then redundant so I'd change it to this

src/synthpop/reproducibility.py
```diff
@@ -1,14 +1,26 @@
...
+class RandomStateManager:
     """
     Manages random numbers and reproducibility in this package.
 
     Instances of this class can be used as a context manager to temporary switch seed:
-    
+
     Examples
     --------
         >>> from reproducibility import RandomStateManager
```
anneliesdej:
```suggestion
        >>> from synthpop.reproducibility import RandomStateManager
```

src/synthpop/reproducibility.py
```diff
@@ -35,64 +47,78 @@ class RandomStateManager():
...
         Set the root seed.
         The intended usage is within the Synthesiser class.
         """
-        cls._root_seed = seed
-        if seed_sequence is None:
-            cls._seed_sequence = SeedSequence(seed)
+
+        # The error messages of SeedSequence when seed is invalid are clear enough.
+
+        if isinstance(seed, (list, tuple, np.ndarray)) and len(seed) == 0:
```
anneliesdej, line 66: I would add a comment on why this part is added when our type hints only allow for integers. Or update the typehints to follow the numpy API

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+    # When: an RNG is created (create_rng)
+    result = RandomStateManager.create_rng(seed=10)
+
+    # The root seed is used to create an RNG
+    patch_default_rng["mock"].assert_called_with([3, 10])
+    assert result is patch_default_rng["expected_rng"], "the returned RNG should be from np.random.default_rng"
+    assert_seed_state(3)
+
+
+def test_random_state_manager_given_uninitialised_when_creating_new_seed(monkeypatch, mocker):
```
anneliesdej: This test needs to be updated (name and docstring) to match the new function name

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+    # Then: The random state manager gets initialised with a random seed.
+    # This asserts that set_root_seed has been called with seed=None.
+    assert_seed_state(expected_seed)
+
+    # Then: A seed is created using the seed sequence
+    mock_create_seed_from_sequence.assert_called_with(expected_seed_sequence)
+    assert result == expected_returned_seed
+
+
+def test_random_state_manager_given_initialised_when_creating_new_seed(monkeypatch, mocker):
```
anneliesdej: This test also needs to be updated with the new function name

tests/integration/test_int_reproducibility.py
```diff
@@ -0,0 +1,96 @@
...
+    result1 = transformer1.transform(X=0)
+    result2 = transformer2.transform(X=0)
+
+    assert not np.array_equal(result1, result2)
+
+    assert np.array_equal(result1, transformer1.transform(X=0))
+    assert np.array_equal(result2, transformer2.transform(X=0))
+
+
+def test_standard_transformer_reproduces_when_setting_seed():
```
anneliesdej, line 84: Maybe a test using sequence seeds is also needed since they are kind of part of the implementation? Like [1, 2, 3]?

anneliesdej (changes requested):

src/synthpop/reproducibility.py
```diff
@@ -1,17 +1,31 @@
+import secrets
+from typing import List
+import warnings
+
 import numpy as np
 from numpy.random import SeedSequence
+import numpy.typing as npt
```
casvanerpduo, line 7: npt is not used in this file. Can be removed? Or used in typehints

src/synthpop/reproducibility.py
```diff
@@ -1,17 +1,31 @@
...
+
+
+def _create_seed_from_sequence(seed_sequence: SeedSequence) -> int:
 
+    # SeedSequence.generate_state returns the same value if you call it multiple times.
+    # To guarantee independent RNGs, we need independent seeds.
+    # That is why we need to generate a new seed sequence for each seed.
+    new_seed_sequence = seed_sequence.spawn(1)[0]
+    new_seed = new_seed_sequence.generate_state(1)[0]  # generate_state generates a list of seeds, we need only one
+    return int(new_seed)
```
casvanerpduo, line 17: Just to think about/consider: I think int() can have an arbitrary number of bits used as memory. Is that intended/usefull? Or would setting new_seed.astype(np.int64/np.int32) be more safe/usefull?

(Also consider that with allocating a specific memory overflows might occur at specific points)

tests/integration/test_int_random_state_manager.py
```diff
@@ -0,0 +1,170 @@
...
+The aim is to test if RandomStateManager behaves as intended.
+
```
casvanerpduo, line 4: Have we taken a look at:
https://en.wikipedia.org/wiki/Randomness_test
https://en.wikipedia.org/wiki/Diehard_tests

Should we include Dieharder test(s)? These test, on a large scale, if the output indeed is random and if our generator is secure.

casvanerpduo: I think the general simplified idea is just to call the RNG function a lot of times and check if there are no correlations, in cases we set seed=None (cause we expect that with seed=number the output is the same, which is tested)

tests/unit/test_random_state_manager.py
```diff
@@ -0,0 +1,347 @@
...
+        monkeypatch.setattr(RandomStateManager,
+                            "_seed_sequence", expected_seed_sequence)
+
+    monkeypatch.setattr(RandomStateManager, "set_root_seed", mock_set_seed)
+
+    expected_returned_seed = 44
+    mock_create_seed_from_sequence = mocker.patch(
+        "synthpop.reproducibility._create_seed_from_sequence", return_value=expected_returned_seed)
+
+    # When: an seed is created (create_new_seed)
```
casvanerpduo:
```suggestion
    # When: a seed is created (create_new_seed)
```

src/synthpop/reproducibility.py
KoomenErina: used in typehints

Issue #134, opened 2026-06-08
Add default values in sphinx build


anneliesdej:
The parameters in readthedocs currently do not show the default values (if applicable). They are only visible in the very first function line. It would be better to change in the sphinx settings that the defaults are more clearly visible by also adding them to the parameter list

PR #151, opened 2026-06-17
145 improve read the docs structure

casvanerpduo:
Made a new Read The Docs structure, similar to the documentation of NumPy and SciPy using the PyData Sphinx theme.

I did not change any of the texts/documentation. For missing text I just left a placeholder.

Please review and leave your thoughts.


tests/unit/__pycache__/test_S_pMSE.cpython-314-pytest-9.0.2.pyc.4296
KoomenErina, line 1: this file should not be included in here

docs/source/examples/examples_index.md
```diff
@@ -0,0 +1,17 @@
+# Examples
+
+This is where we would explain examples, etc.
+
+```{toctree}
+:maxdepth: 1
+
+regular_use.md
```
KoomenErina, line 16: The rendered version shows a list of links to each individual example. Is it possible to include some preview/explanation for each link?

KoomenErina:
I find the function signatures of for example `CartSynthesis.transform` not that easy to read. Is there a way to reduce the amount of information there?

KoomenErina:
The api reference of reproducibility is in the section called "Synthesiser class".  That seems misleading to me

KoomenErina:
There needs to be a way to mark certain classes as internal/unstable.

KoomenErina:
There are a few human targeted documentation files that are required/suggested by the standards of public code:
- CONTRIBUTING.md
- GOVERNANCE.md
- README.md
- LICENSE.md
- SECURITY.md

Each of these are expected to be in the root of the repository. GitHub expects those files there and integrates it in the only UI ( the License symbol with the scales icon, for example). 

I think that some of these files should/can be part of the readthedocs. Is there a way to include those files?

docs/source/examples/examples_index.md
anneliesdej, line 16: I would split the examples more in topics, like sklearn does. You would probably get something like the basics, parameters (which includes the column order), synthesis methods, custom synthesis

docs/source/installing/installing.md
```diff
@@ -0,0 +1,4 @@
+# Installing
```
anneliesdej: I would change the name installing to installation (numpy uses this) or install (scikit-learn uses this)

docs/source/developer/functional_descriptions/fd.md
```diff
@@ -1,5 +1,3 @@
-# Functional descriptions
-This documentation is used by developers to implement this package. It can be very detailed. 
 ```{toctree}
```
anneliesdej: I think here, you might want to have more pages hidden. The side bar is now very full with all the FDs. Or a clearer structure. While the order makes sense if you know the package, it is pretty long. So you probably would like the same subcategories as in the API reference

docs/source/developer/way_of_working/randomness.md
```diff
@@ -1,6 +1,6 @@
 # Using randomness in this package
 
-This package enforces strict requirements on randomness to ensure both reproducibility (deterministic outputs given a seed) and statistical validaty (high-quality, reproducible random streams). These requirements are defined in the [functional descriptions](../functional%20descriptions/reproducibility.md) and are consistent with guidance from the
+This package enforces strict requirements on randomness to ensure both reproducibility (deterministic outputs given a seed) and statistical validaty (high-quality, reproducible random streams). These requirements are defined in the [functional descriptions](../functional_descriptions/reproducibility.md) and are consistent with guidance from the
 [scikit-learn's](https://scikit-learn.org/stable/developers/develop.html#random-numbers) standards for random numbers and aligned with [NumPy's](https://numpy.org/doc/stable/reference/random/generator.html) Generator API and [seeding recommendations](https://numpy.org/doc/stable/reference/random/bit_generators/index.html#seeding-and-entropy), with a focus on deterministic reproducibility.
 
 ## Background: NumPy vs. scikitlearn
```
anneliesdej:
```suggestion
## Background: NumPy vs. scikit-learn
```

docs/source/developer/way_of_working/way_of_working.md
```diff
@@ -0,0 +1,11 @@
+```{toctree}
```
anneliesdej: This page has no title rendered, and you might not even want this page to show as the index is already shown on the page before

docs/source/developer/functional_descriptions/fd.md
```diff
@@ -1,5 +1,3 @@
...
 :maxdepth: 1
```
anneliesdej: This page also has no title so you get:
for developers > <no title> > synthpop synthesis

docs/source/index.rst
```diff
@@ -15,12 +17,14 @@ documentation for details.
    :maxdepth: 1
    :caption: Contents:
 
-   intro
-   synthpop
-   user guides/examples
-   functional descriptions/fd_index
+   installing/installing.md
```
anneliesdej, line 26: Would Install, User Guides, API, Examples be more logical? That is the order that scikit-learn uses.

anneliesdej (changes requested):
I'm a big fan of the new structure. I suggest some small tweaks

docs/source/examples/examples_index.md
casvanerpduo: With the :hidden: argument we can hide the list, and write a text in the examples_index.md using links to the files. I added placeholders for this.

docs/source/developer/functional_descriptions/fd.md
casvanerpduo: I think it is better to keep, as long as possible, every subcategory clickable from a single page. This makes sure that a user can easily find what it needs. 

It is a good idea to reduce the amount of functional descriptions by splitting it in different subjects. I think the best way to proceed is by putting it all in the same file, than a clickable structure/tree appears to the right hand side(similar to API Reference-CART Method).

I think the latter is an idea for a different github issue, that would tackle the exact files and substructures.

docs/source/about_us/about.md
```diff
@@ -1,2 +1,6 @@
 # About us
```
KoomenErina, line 1: Is it possible that the about-us part is build-up out of multiple sections like in examples_index.md? That way we can put the licence, governance, security as sections and perhaps the readme as main-page of 'about`

I think a part of the content of the readme belongs in the about-us page

docs/source/developer/developer_index.md
```diff
@@ -12,23 +12,54 @@ For questions about what constitutes acceptable code, consult [Code Standards, N
 
 Finally, the [Dataflow Diagrams](way_of_working/dataflowdiagram.md) provide a visual and conceptual overview of how Synthpop operates internally. Use this as a reference when you need to understand the system's architecture or the flow of data through different components.
 
+```{include} ../../../GOVERNANCE.md
```
KoomenErina: I would say that governance.md should not be in the developer docs, but in the about-us section

docs/source/conf.py
```diff
@@ -46,17 +46,18 @@
...
+    'sphinx.ext.napoleon',
     'myst_parser',
-    "sphinxcontrib.mermaid",
+    'sphinxcontrib.mermaid',
 ]
 myst_enable_extensions = ["dollarmath", "amsmath","tasklist"]
-autodoc_typehints = "both"
 autodoc_member_order = "bysource"
 
+autodoc_typehints = "none"
```
KoomenErina: The typehints are not shown in the docs. Perhaps show the typehints in the parameter list, but not in the function signature? (https://pypi.org/project/sphinx-autodoc-typehints/#user-content-how-to-guides )

KoomenErina: Is this a good point to handle #134 ?

pyproject.toml
```diff
@@ -20,7 +20,8 @@ dependencies = [
 ]
 
 [dependency-groups]
-docs = [ 
+docs = [
+    "pydata-sphinx-theme (>=0.19.0,<1.0.0)",
     "sphinx-rtd-theme (>=3.0.2,<4.0.0)",
```
KoomenErina: we don't need the sphinx theme if we use the pydata theme

docs/source/conf.py
casvanerpduo: This can be done by setting autodoc_tyhpehints to description. I tried this but found the formatting of the typehints hard to interpret. This probably is, as far as I understood, due to the formatting of typehints. Other packages do not use typehints this way, and provide information on datatyping in their functional descriptions.

As such I chose forautodoc_typheints = none. Changing the formatting would require changing the docstrings to be more similar to, for instance:
https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/cluster/_agglomerative.py
https://github.com/numpy/numpy/blob/main/numpy/linalg/_linalg.py

I have set it to description for now.

casvanerpduo: Perhaps together with the point above

docs/source/about_us/about.md
casvanerpduo: Yes! I have now added Governance and License in similar method. Basically one would need to define it's own toctree for a different section.

docs/source/developer/functional_descriptions/fd.md
anneliesdej: Agreed, I will create an issue

docs/source/developer/developer_index.md
```diff
@@ -1,25 +1,59 @@
...
 
-For questions about what constitutes acceptable code, consult [Code Standards, Norms, and Conventions](code_standards_and_norms.md). This document defines the criteria for code quality and consistency, and acts as a tie-breaker when multiple valid approaches exist.
+For questions about what constitutes acceptable code, consult [Code Standards, Norms, and Conventions](way_of_working/code_standards_and_norms.md). This document defines the criteria for code quality and consistency, and acts as a tie-breaker when multiple valid approaches exist.
 
-Finally, the [Dataflow Diagrams](dataflowdiagram.md) provide a visual and conceptual overview of how Synthpop operates internally. Use this as a reference when you need to understand the system's architecture or the flow of data through different components.
+Finally, the [Dataflow Diagrams](way_of_working/dataflowdiagram.md) provide a visual and conceptual overview of how Synthpop operates internally. Use this as a reference when you need to understand the system's architecture or the flow of data through different components.
+
+```{include} ../../../GOVERNANCE.md
```
KoomenErina: Governance.md should not be included here

LICENSE.md
```diff
@@ -1,3 +1,4 @@
+# License
```
KoomenErina: I suspect that this file is going to be read by machines a lot, especially the first few lines. I think that it is best not to modify this file.

PR #157, opened 2026-06-23
147 make the synthesis process reproducible

KoomenErina:
This pull requests makes the entire proces of making synthetic data reproducible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 classes.dot
 pyproject.toml
 tests/unit/__pycache__/
+*.png
+*.svg
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ pyproject.toml
 tests/unit/__pycache__/
 *.png
 *.svg
+tests/integration/__pycache__/
--- a/src/synthpop/data_processing/missing_value_handling.py
+++ b/src/synthpop/data_processing/missing_value_handling.py
@@ -13,6 +13,7 @@ from sklearn.exceptions import NotFittedError
 
 from synthpop.data_processing.encoders import MeanEncoder
 from synthpop.methods.tree_utils import LeafNodeSampler, build_feature_matrix
+from synthpop.reproducibility import RandomStateManager
 from synthpop.utils import validate_2d_dict, validate_1d_target
 
 class BaseMissingValueHandler(metaclass=ABCMeta):
@@ -140,7 +141,7 @@ class MissingValuePredictor(BaseMissingValueHandler):
 
         self.feature_order_ = list(X_val.keys())
 
-        self.tree_ = clone(self.tree) if self.tree else DecisionTreeClassifier(min_samples_leaf=5)
+        self.tree_ = clone(self.tree) if self.tree else DecisionTreeClassifier(min_samples_leaf=5,random_state=RandomStateManager.create_instance_seed())
         self.tree_sampler_ = self.tree_sampler.clone() if self.tree_sampler else LeafNodeSampler()
         
         # implementation
--- a/src/synthpop/methods/cart_synth.py
+++ b/src/synthpop/methods/cart_synth.py
@@ -22,6 +22,7 @@ from synthpop.data_processing.missing_value_handling import (
 )
 from synthpop.methods import base_synth
 from synthpop.methods.tree_utils import LeafNodeSampler
+from synthpop.reproducibility import RandomStateManager
 
 
 def _to_fixed_length_string_array(a: npt.NDArray) -> npt.NDArray:
@@ -232,8 +233,9 @@ class TreeClassifierMethod(_AbstractTreeMethod):
 
     def _get_tree(self):
         return DecisionTreeClassifier(min_samples_leaf=5,   # equivalent to minbucket in synthpop-r
-                                      min_impurity_decrease=1e-08  # equivalent to cp in synthpop-r
-                                      ,)
+                                      min_impurity_decrease=1e-08,  # equivalent to cp in synthpop-r
+                                      random_state=RandomStateManager.create_instance_seed(),
+                                      )
 
     def _convert_y(self, y: npt.NDArray) -> npt.NDArray:
         return _to_fixed_length_string_array(y)
@@ -292,6 +294,7 @@ class TreeRegressorMethod(_AbstractTreeMethod):
     def _get_tree(self):
         return DecisionTreeRegressor(min_samples_leaf=5,    # equivalent to minbucket in synthpop-r
                                     min_impurity_decrease= 1e-08,   # equivalent to cp in synthpop-r
+                                    random_state=RandomStateManager.create_instance_seed()
                                     )
     
     def _convert_y(self, y: npt.NDArray) -> npt.NDArray:
--- a/src/synthpop/methods/cart_synth.py
+++ b/src/synthpop/methods/cart_synth.py
@@ -22,6 +22,7 @@ from synthpop.data_processing.missing_value_handling import (
 )
 from synthpop.methods import base_synth
 from synthpop.methods.tree_utils import LeafNodeSampler
+from synthpop.reproducibility import RandomStateManager
 
 
 def _to_fixed_length_string_array(a: npt.NDArray) -> npt.NDArray:
@@ -71,6 +72,46 @@ class _AbstractTreeMethod(TransformerMixin, BaseEstimator, metaclass=ABCMeta):
     def _convert_y(self, y: npt.NDArray) -> npt.NDArray:
         # overwritten in TreeClassifierMethod and TreeRegressorMethod
         return y
+    
+    def print_decision_path(self,X_test,sample_id = 0):
+        clf = self.tree_
+        node_indicator = clf.decision_path(X_test)
+        leaf_id = clf.apply(X_test)
+
+        
+        feature= feature = clf.tree_.feature
+        threshold = clf.tree_.threshold
+        values = clf.tree_.value
+        # obtain ids of the nodes `sample_id` goes through, i.e., row `sample_id`
+        node_index = node_indicator.indices[
+            node_indicator.indptr[sample_id] : node_indicator.indptr[sample_id + 1]
+        ]
+
+        print("Rules used to predict sample {id}:\n".format(id=sample_id))
+        for node_id in node_index:
+            # continue to the next node if it is a leaf node
+            if leaf_id[sample_id] == node_id:
+                continue
+
+            # check if value of the split feature for sample 0 is below threshold
+            if X_test[sample_id, feature[node_id]] <= threshold[node_id]:
+                threshold_sign = "<="
+            else:
+                threshold_sign = ">"
+
+            print(
+                "decision node {node} : (X_test[{sample}, {feature}] = {value}) "
+                "{inequality} {threshold}), missing go left: {missing_go_left}".format(
+                    node=node_id,
+                    sample=sample_id,
+                    feature=feature[node_id],
+                    value=X_test[sample_id, feature[node_id]],
+                    inequality=threshold_sign,
+                    threshold=threshold[node_id],
+                    missing_go_left=clf.tree_.missing_go_to_left[node_id]
+                )
+            )
+
 
     def fit(self, X: Dict[str, npt.NDArray], y: npt.NDArray) -> Self:
         """
@@ -104,6 +145,9 @@ class _AbstractTreeMethod(TransformerMixin, BaseEstimator, metaclass=ABCMeta):
 
         leaf_ids = self.tree_.apply(all_features)
 
+        if(self.n_features_in_ == 100):
+            print(self.n_features_in_)
+
         self.tree_sampler_ = self._new_tree_sampler().fit_sampler(leaf_ids, prepared_y)
 
         return self
@@ -133,6 +177,9 @@ class _AbstractTreeMethod(TransformerMixin, BaseEstimator, metaclass=ABCMeta):
         X_val, _ = utils.validate_2d_dict(X)
 
         n_features_given = len(X.keys())
+
+        if(n_features_given == 100):
+            print(n_features_given)
         if n_features_given != self.n_features_in_:
             raise ValueError(
                 f"X has {n_features_given} features, but {self.__class__.__name__} is expecting {self.n_features_in_} features as input")
@@ -232,7 +279,8 @@ class TreeClassifierMethod(_AbstractTreeMethod):
 
     def _get_tree(self):
         return DecisionTreeClassifier(min_samples_leaf=5,   # equivalent to minbucket in synthpop-r
-                                      min_impurity_decrease=1e-08  # equivalent to cp in synthpop-r
+                                      min_impurity_decrease=1e-08,  # equivalent to cp in synthpop-r
+                                      random_state=RandomStateManager.create_new_seed()
                                       ,)
 
     def _convert_y(self, y: npt.NDArray) -> npt.NDArray:
@@ -292,6 +340,7 @@ class TreeRegressorMethod(_AbstractTreeMethod):
     def _get_tree(self):
         return DecisionTreeRegressor(min_samples_leaf=5,    # equivalent to minbucket in synthpop-r
                                     min_impurity_decrease= 1e-08,   # equivalent to cp in synthpop-r
+                                    random_state=RandomStateManager.create_new_seed()
                                     )
     
     def _convert_y(self, y: npt.NDArray) -> npt.NDArray:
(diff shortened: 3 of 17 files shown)


src/synthpop/methods/tree_utils.py
```diff
@@ -122,13 +124,16 @@ def fit_sampler(self, leaf_ids: npt.ArrayLike, y: npt.ArrayLike) -> Self:
                     "distributions and may be sampled. Review your input data "
                     "if this is unintended.")
 
-        if isinstance(self.random_state, np.random.Generator):
-            # Extract seed is not possible → treat as non-resettable (repeated calls do not give the same output)
-            self._seed = None
-            self.random_state_ = self.random_state
+        # if isinstance(self.random_state, np.random.Generator):
```
anneliesdej: Can these comments be removed or do you think they still add necessary information?

src/synthpop/methods/tree_utils.py
```diff
@@ -159,12 +164,14 @@ def sample_from_leaves(self, leaf_ids: npt.ArrayLike) -> npt.NDArray:
...
-        if seed is not None:
-            rng = np.random.default_rng(seed)
-        else:
-            rng = self.random_state_ # fallback, not-resettable
+        # if seed is not None:
+        #     rng = np.random.default_rng(seed)
+        # else:
+        #     rng = self.random_state_ # fallback, not-resettable
+
+        rng = RandomStateManager.create_rng(self.random_state_)
```
anneliesdej, line 165: I would add some information about the randomness in sample_from_leaves specifically. Every call resets RNG state to the initial seed state. This means that
```python
sampler.sample_from_leaves(A)
sampler.sample_from_leaves(A)
```
Gives the same results. This is what we want to have, but it is not as common. Evolving randomness is more common than replayable randomness.

It might also be enough to make sure this is tested, but I think it is helpful to add as a comment here.

src/synthpop/methods/cart_synth.py
anneliesdej, line 228: The default svd_solver of PCA() is 'auto' and it states this:
"The solver is selected by a default ‘auto’ policy is based on X.shape and n_components: if the input data has fewer than 1000 features and more than 10 times as many samples, then the “covariance_eigh” solver is used. Otherwise, if the input data is larger than 500x500 and the number of components to extract is lower than 80% of the smallest dimension of the data, then the more efficient “randomized” method is selected. Otherwise the exact “full” SVD is computed and optionally truncated afterwards."

Which means that randomness can be introduced dependent on the input data. We could force the solver to use 'full', or add a random_state there (probably the preferred option)

src/synthpop/methods/sample_synth.py
anneliesdej:
```suggestion
```
np is not used

src/synthpop/methods/sample_synth.py
```diff
@@ -92,8 +95,9 @@ def transform(self, X: pd.DataFrame | None) -> pd.Series:
             raise NotFittedError("SampleMethod is not fitted. Call `fit` first.")
         
         n = len(X) if X is not None else self.n_samples_
+        rng = RandomStateManager.create_rng(seed=self.random_state_)
```
anneliesdej, line 97: Maybe also add here that it is a replay RNG and not a streaming RNG

tests/unit/test_sample_method.py
```diff
@@ -92,6 +101,24 @@ def test_transform_values_within_observed_support():
     if any(pd.isna(v) for v in values):
         assert result.isna().any()
 
+def test_transform_uses_random_state_manager(mocker):
```
anneliesdej, line 119: Integration tests need to be added to test reproducibility. Same seed gives same results and repeated calls give same results

src/synthpop/methods/sample_synth.py
KoomenErina: Writing in the docs of `create_rng` that it is a replay RNG would be more efficient I think

src/synthpop/methods/tree_utils.py
KoomenErina: added it to the doc string

src/synthpop/methods/sample_synth.py
anneliesdej: Yes, agreed

tests/unit/test_pca_encoder.py
```diff
@@ -11,9 +11,10 @@
 
 class TransformStub(TransformerMixin, BaseEstimator):
 
-    def __init__(self, fit_return_value=None, transform_return_value=None):
+    def __init__(self, fit_return_value=None, transform_return_value=None,random_state=None):
```
anneliesdej, line 14: this file needs an autopep8

tests/unit/test_sample_method.py
```diff
@@ -31,19 +31,28 @@ def make_fitted_model(values: list, counts: list, target_name="target", n_sample
         (pd.Series([np.nan, np.nan, np.nan, np.nan, 0], name="my_target"), [np.nan, 0], [4, 1],),
     ],
 )
-def test_fit_stores_distribution_and_metadata(y, expected_values, expected_counts):
+def test_fit_stores_distribution_and_metadata(y, expected_values, expected_counts,mocker):
```
anneliesdej: This file needs an autopep8

tests/unit/test_synthesiser.py
```diff
@@ -174,6 +175,81 @@ def test_synthesiser_fit_custom_order_by_column_index():
...
+
+        def fit(self,X,y):
+            assert MockRandomStateManager.is_in_contextblock, "fit called outside context block"
+            return super().fit(X,y)
+        
+        def transform(self,X):
+            assert MockRandomStateManager.is_in_contextblock, "transform called outside context block"
+            return super().transform(X)
+
+    return mocker.patch(
```
anneliesdej, line 216: This file needs an autopep8

.gitignore
```diff
@@ -10,5 +10,6 @@
 classes.dot
 pyproject.toml
 tests/unit/__pycache__/
+*.png
```
anneliesdej:
```suggestion
```
Is already in the list two lines below

tests/unit/test_leaf_node_sampler.py
```diff
@@ -159,9 +182,12 @@ def integers(self, low, high, size=None):
             out.append(low + (v % (high - low)))
         return np.array(out)
     
-def test_sampling_deterministic_with_stub_rng():
+def test_sampling_deterministic_with_stub_rng(mocker):
+    #TODO: inject the stub by mocking RandomStateManager.create_rng
```
anneliesdej: Still a todo

tests/unit/test_leaf_node_sampler.py
```diff
@@ -194,6 +220,7 @@ def test_sample_from_leaves_each_missing_type(missing_value):
         assert any(isinstance(v, float) and np.isnan(v) for v in y_syn)
 
 def test_sample_determinism_with_same_seed():
+    #TODO: move to integration test
```
anneliesdej: still a todo

tests/unit/test_sample_method.py
```diff
@@ -92,6 +101,24 @@ def test_transform_values_within_observed_support():
...
+
+    result = model.transform(pd.DataFrame(index=range(100)))
+
+    mock_create_rng.assert_called_with(seed=123456)
+    mock_sample_array.assert_called_with(expected_rng,model.counts_,model.values_,100)
+    assert (result==expected_result).all()
+
+
 
 def test_transform_reproducibility_with_fixed_seed():
```
anneliesdej, line 140: Here I miss a test that checks what happens with two calls (if it is streaming or replayable)

tests/integration/test_int_reproducibility.py
```diff
@@ -96,3 +99,78 @@ def test_standard_transformer_reproduces_when_setting_seed():
     result2 = transformer1.transform(X=2)
```
anneliesdej:
```suggestion
    result2 = transformer2.transform(X=2)
```
I think this should be transfomer2

tests/integration/test_int_reproducibility.py
```diff
@@ -6,8 +6,11 @@
 import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
 
+from synthpop.methods.sample_synth import SampleMethod
 from synthpop.reproducibility import RandomStateManager
-
+import pandas as pd
+from synthpop.synthesiser import Synthesiser
+from tests.integration.data_generated_for_tests import get_test_data_classifier,get_test_data_regressor,simulate_realistic_dataset_correlations
```
anneliesdej, line 25: Should the sample method also be added here or are unit tests enough?

tests/integration/test_int_reproducibility.py
```diff
@@ -6,8 +6,11 @@
...
 
 @pytest.fixture(autouse=True)
 def control_random_state_manager():
```
anneliesdej, line 27: Maybe add one line of explanation that we do not use the big fixture as this test needs to set reproducibility and not use the default of the big fixture (in make_int_test_reproducible)

tests/integration/test_int_reproducibility.py
```diff
@@ -96,3 +99,78 @@ def test_standard_transformer_reproduces_when_setting_seed():
...
+        else:
+            d_data[k] = X_clas[k]
+
+    d_data['y1'] = y_reg
+    d_data['y2']= y_clas
+
+    return pd.DataFrame(d_data)
+
+
+
```
anneliesdej: I would add one more test for completeness. Right now we verified that
```python
fit(seed)
generate()
generate()
```
is replayable, but I would also add a test that overrides the seed in generate to check if that is also replayable. So something like:
```python
fit(seed)
generate(other seed)
generate(same other seed)
```

tests/integration/test_int_reproducibility.py
KoomenErina: see `test_sample_method_reproducible`, it is already here

tests/unit/test_sample_method.py
KoomenErina: see `test_sample_method_reproducible` in test_int_reproducibility.py

tests/unit/test_leaf_node_sampler.py
```diff
@@ -106,14 +106,37 @@ def test_fit_sampler_sets_y_dtype_correctly(y_input, expected_dtype_check):
 
     assert np.issubdtype(sampler._y_dtype, expected_dtype_check)
 
+def test_fit_sampler_creates_seed_from_random_state_manager(mocker):
```
anneliesdej, line 115: This file would benefit from autopep8

tests/integration/test_int_tree_method.py
```diff
@@ -10,9 +10,14 @@
 from synthpop.data_processing.encoders import PCAEncoder
 from synthpop.data_processing.missing_value_handling import MissingValuePredictor
 from synthpop.methods.cart_synth import TreeClassifierMethod, TreeRegressorMethod
+from synthpop.reproducibility import RandomStateManager
 from synthpop.utils import str_dtype
 from sklearn.datasets import make_classification, make_regression
```
anneliesdej, line 15: This file would benefit from autopep8

tests/integration/test_int_reproducibility.py
```diff
@@ -93,6 +103,119 @@ def test_standard_transformer_reproduces_when_setting_seed():
...
+def combined_regressor_and_classifier_test_data(seed=10):
+    X_reg, y_reg = get_test_data_regressor(
+        seed=seed, with_cats=True, with_missing_features=True, with_missing_target=True)
+    X_clas, y_clas = get_test_data_classifier(
+        seed=seed, with_cats=True, with_missing_features=True, with_missing_target=True)
+
+    d_data = {}
+
+    available_columns = set(X_reg.keys()).intersection(set(X_clas.keys()))
```
anneliesdej:
```suggestion
    available_columns = sorted(set(X_reg) & set(X_clas))
```
The helper currently is technically non-deterministic as python sets are unordered. Since it is used for reproducibility tests, I'd make it deterministic

tests/integration/test_int_reproducibility.py
```diff
@@ -93,6 +103,119 @@ def test_standard_transformer_reproduces_when_setting_seed():
...
+
+    syn1 = synth.generate()
+    syn2 = synth.generate(random_seed=100)
+    syn3 = synth.generate(random_seed=100)
+
+    assert syn2.equals(syn3)
+    assert not syn2.equals(syn1)
+
+
+def test_sample_method_reproducible():
```
anneliesdej, line 201: For completeness I would not only check if transform is replayable but also if a seed is reproducible (like the other tests have):
```python
def test_sample_method_explicit_seed_reproducible():

    y = ...

    m1 = SampleMethod(random_state=123).fit(None, y)
    m2 = SampleMethod(random_state=123).fit(None, y)

    assert m1.transform(None).equals(m2.transform(None))
```

tests/integration/test_int_reproducibility.py
```diff
@@ -93,6 +103,119 @@ def test_standard_transformer_reproduces_when_setting_seed():
...
+    syn2 = synth.generate(n=100, random_seed=1234)
+
+    assert not syn1.equals(syn2)
+
+    synth2 = Synthesiser(random_seed=1234)
+    syn3 = synth2.fit(obs).generate(n=100)
+    assert not syn1.equals(syn3)
+
+
+def test_generate_independent_syn_datasets_reproducible():
```
anneliesdej:
```suggestion
def test_generate_override_seed_is_replayable():
```
This describes the test better

src/synthpop/data_processing/missing_value_handling.py
```diff
@@ -140,7 +141,7 @@ def prepare_data_for_fit(self, X: Dict[str, npt.NDArray], y: npt.NDArray) -> tup
 
         self.feature_order_ = list(X_val.keys())
```
anneliesdej, line 143: This file would benefit from autopep8

src/synthpop/methods/tree_utils.py
```diff
@@ -159,12 +161,14 @@ def sample_from_leaves(self, leaf_ids: npt.ArrayLike) -> npt.NDArray:
         ):
             raise NotFittedError("LeafNodeSampler is not fitted. Call `fit_sampler` first.")
         
-        seed = getattr(self, "_seed", None)
+        # seed = getattr(self, "_seed", None)
```
anneliesdej: I think these comment lines can be removed

src/synthpop/reproducibility.py
casvanerpduo, line 28: Also add an Example on how to use this with the Synthesizer class and synthesize data. That is now unclear to me.

Perhaps also extend the doc strings to also explain how it manages reproducibility, and why that is required

KoomenErina: The user does not use this class directly when using the Synthesiser class. Examples of reproducible synthesis are in the example docs

KoomenErina: The functional description (reproducibility.md) explains the why, and for a part the how. The developerguide (randomness.md) explains it more practically

tests/integration/data_generated_for_tests.py
```diff
@@ -0,0 +1,111 @@
+import copy
+import pytest
+import pandas as pd
+import numpy as np
+import string
+
+from synthpop.utils import str_dtype
+from sklearn.datasets import make_classification, make_regression
```
casvanerpduo:
```suggestion
import copy
import string

import pytest
import pandas as pd
import numpy as np
from sklearn.datasets import make_classification, make_regression

from synthpop.utils import str_dtype
```

tests/integration/test_int_reproducibility.py
```diff
@@ -6,11 +6,21 @@
 import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
 
+from synthpop.methods.sample_synth import SampleMethod
+from synthpop.methods.tree_utils import LeafNodeSampler
 from synthpop.reproducibility import RandomStateManager
+import pandas as pd
+from synthpop.synthesiser import Synthesiser
+from tests.integration.data_generated_for_tests import get_test_data_classifier, get_test_data_regressor, simulate_realistic_dataset_correlations
```
casvanerpduo, line 23:
```suggestion
import pandas as pd
import pytest
import numpy as np
from sklearn.base import BaseEstimator, TransformerMixin

from synthpop.methods.sample_synth import SampleMethod
from synthpop.methods.tree_utils import LeafNodeSampler
from synthpop.reproducibility import RandomStateManager
from synthpop.synthesiser import Synthesiser
from tests.integration.data_generated_for_tests import get_test_data_classifier, get_test_data_regressor, simulate_realistic_dataset_correlations
```

PR #159, opened 2026-06-23
153 runtimewarning in pairwise spmse

casvanerpduo:
The problem regarding this runtimewarning is a datatype error. The S_pMSE did not handle different datatype inputs correctly, in cases where the column names of the synthetic and original dataset were the same. This produced two problems:

1. In cases where the original dataset is numeric, and the synthetic is not. pd.cut produces an error
2.  calculating the rescaled differences and expected frequencies produces a runtimewarning, and creates invalid s_pmse values

I added a ValueError raise, including a unit-test, that checks on the datatypes of the columns.

### DIFF
--- a/src/synthpop/utility_metrics/spmse.py
+++ b/src/synthpop/utility_metrics/spmse.py
@@ -7,6 +7,7 @@ import warnings
 
 import numpy as np
 import pandas as pd
+from synthpop.utils import standardise_array_dtypes
 
 __all__ = ["pairwise_spmse"]
 
@@ -160,13 +161,6 @@ def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int = 
         raise ValueError(
             "Original and synthetic dataframes must have the same shape and column names.")
 
-    # Requires ordering
-    datatype_check = (orig_df.dtypes == syn_df[orig_df.columns].dtypes)
-    if not datatype_check.all():
-        raise ValueError(
-            f"Original and synthetic dataframes must have the same datatypes. " +
-            f"Error is in columns {(orig_df.columns[~datatype_check]).tolist()}.")
-
     if max_bins < 1 or not isinstance(max_bins, int):
         raise ValueError(
             "The number of bins should be an integer with value of at least 1.")
@@ -176,8 +170,11 @@ def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int = 
     s_df = syn_df.copy(deep=False)
 
     # Start calculations for preprocessing here
-    for column_name in orig_df:
+    for column_name in o_df:
 
+        o_df[column_name] = standardise_array_dtypes(o_df[column_name])
+        s_df[column_name] = standardise_array_dtypes(s_df[column_name])
+        
         if pd.api.types.is_numeric_dtype(o_df[column_name]):
             combined = pd.concat([o_df[column_name], s_df[column_name]])
             _, bins = pd.cut(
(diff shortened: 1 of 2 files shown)


tests/unit/test_S_pMSE.py
```diff
@@ -50,6 +50,13 @@
         ),
         # Check empty DataFrames
 
+        (
+            pd.DataFrame({"c1": [0, 0, 0, 1]}, dtype='category'),
+            pd.DataFrame({"c1": ['0', '1']}, dtype='string'),
+            25,
+            f"['c1']."
```
KoomenErina: the argument of `match` is seen as a regex expression. square brackets ('[ ]`) have meaning in regex. Perhaps use `re.escape`?

KoomenErina (changes requested):

KoomenErina:
The warning itself mentions to set `sort=False` to suppres this warning. Is there a reason why that is not sufficient?

Does a mismatch in datatypes causes incorrect output?

KoomenErina:
The current version of the synthesis process does not preserve dtypes in all situations. Raising an error when the dtypes do not match would prevent the user to evaluate the synthesis with regular usage. 

The error mentioned that "the values are unorderable". What happens if both dtypes are `object`?

I think that the output dytpes of the synthesis process is something we should look at. At the same time, if `pairwise_spmse` produces correct outputs for unorderable/different dtypes, there is no need to raise an error.

anneliesdej (changes requested):
A regression test is also needed if possible

src/synthpop/utility_metrics/spmse.py
```diff
@@ -195,6 +200,11 @@ def pairwise_spmse(orig_df: pd.DataFrame, syn_df: pd.DataFrame, max_bins: int =
 
         jf_orig = _joint_frequencies(o_df, col1, col2)
         jf_syn = _joint_frequencies(s_df, col1, col2)
+
+        full_index = jf_orig.index.union(jf_syn.index, sort=False)
```
KoomenErina, line 210: Why is this needed? (line 204 to 206)

tests/unit/test_S_pMSE.py
```diff
@@ -167,6 +166,55 @@ def test_pairwise_spmse_raises_wrong_inputs(orig_df, syn_df, max_bins, error):
...
+            pd.DataFrame({"c1": ['a', 'b', 'c']}, dtype='category'),
+            pd.DataFrame({"c1": ['b', 'c']}, dtype='string'),
+            pd.DataFrame(
+                {
+                    "column1": ["c1"],
+                    "column2": ["c1"],
+                    "S_pMSE": [50/72]
+                }
+            ),
+        ),  # Case where the synthesised data has a different datatype
```
KoomenErina, line 201: It might be good to mention the bug issue in the test case that reproduced the bug

src/synthpop/utility_metrics/spmse.py
casvanerpduo: pd.DataFrame.sub() wants to subtract columns with different row names/indices
This occurs at instances where some combinations are present in the original data, and some are (not) in the synthetic data. Inherently, it was thought it wouldn't matter due to fill_value=0 in sub/add. But this solely fixes the case for nan.

The behaviour of the function does not change. Pandas inherently turns to sort=False. but raises a warning when it can't sort and it is not set explicitly (standard value is None)

Now by setting the index.union in rows 203-205. We manually take care of the warning. As both jf_syn/jf_or will have the same indices

KoomenErina (approved):

PR #161, opened 2026-06-23
152 bug cart method does not work with nan array

casvanerpduo:
First version to create a work around for an entire nan array input in CART Method.

I added an attribute: self.method_._all_missing for continuity with BaseMissingValueHandler, which also stores this attribute. Please be very critical on whether this is a good idea.

There is also two tests included that particularly checks whether the fitting is skipped in cart.fit, and one that checks whether the transformation returns the desired output. 

Important to note is that there was a little work-around required due to the check_is_fitted(), as that raises errors.

I also added self._all_missing = False to the StubClassifier/Regressor classes for testing, else the code breaks for the stubs not having the attribute.

### DIFF
--- a/src/synthpop/methods/cart_synth.py
+++ b/src/synthpop/methods/cart_synth.py
@@ -11,6 +11,7 @@ from sklearn import clone
 from sklearn.base import BaseEstimator, TransformerMixin, check_is_fitted
 from sklearn.decomposition import PCA
 from sklearn.tree import BaseDecisionTree, DecisionTreeClassifier, DecisionTreeRegressor
+from sklearn.exceptions import NotFittedError
 
 from synthpop import utils
 import synthpop.methods.tree_utils as tree_utils
@@ -86,12 +87,12 @@ class _AbstractTreeMethod(TransformerMixin, BaseEstimator, metaclass=ABCMeta):
         y = utils.validate_1d_target(y, n_samples)
         self._all_missing = False
 
+        self.n_features_in_ = len(X.keys())
+        self.feature_order_ = list(X.keys())
+
         if pd.isna(y).all():
             self._all_missing = True
             return self
-
-        self.n_features_in_ = len(X.keys())
-        self.feature_order_ = list(X.keys())
 
         self.encoders_ = {name: self._new_encoder().fit(value, y) for (
             name, value) in X_val.items() if not pd.api.types.is_numeric_dtype(value.dtype)}
@@ -123,6 +124,12 @@ class _AbstractTreeMethod(TransformerMixin, BaseEstimator, metaclass=ABCMeta):
 
         """
 
+        if not hasattr(self, '_all_missing'):
+            raise NotFittedError
+        elif self._all_missing:
+            return np.array([np.nan]*len(X[next(iter(X))]))
+
+        
         # Apply encoding, sample, apply (inverse) handling of missing values.
         check_is_fitted(
             self, 
@@ -418,13 +425,6 @@ class CartMethod(base_synth.BaseSynthMethod):
         :param X: Feature dataset.
         :return: Synthesised target variable.
         """
-
-        if hasattr(self, "method_") and getattr(self.method_, "_all_missing", False):
-            return pd.Series(
-            np.nan,
-            index=X.index,
-            name=self.target_name_,
-        )
 
         check_is_fitted(self, ["method_", "feature_names_in_", "target_name_"])
 
--- a/src/synthpop/methods/cart_synth.py
+++ b/src/synthpop/methods/cart_synth.py
@@ -419,7 +419,7 @@ class CartMethod(base_synth.BaseSynthMethod):
         :return: Synthesised target variable.
         """
 
-        if self.method_._all_missing:
+        if hasattr(self, "method_") and getattr(self.method_, "_all_missing", False):
             return pd.Series(
             np.nan,
             index=X.index,
--- a/src/synthpop/methods/cart_synth.py
+++ b/src/synthpop/methods/cart_synth.py
@@ -84,6 +84,11 @@ class _AbstractTreeMethod(TransformerMixin, BaseEstimator, metaclass=ABCMeta):
         self.target_name_ = getattr(y, "name", None)
         X_val, n_samples = utils.validate_2d_dict(X)
         y = utils.validate_1d_target(y, n_samples)
+        self._all_missing = False
+
+        if pd.isna(y).all():
+            self._all_missing = True
+            return self
 
         self.n_features_in_ = len(X.keys())
         self.feature_order_ = list(X.keys())
@@ -413,6 +418,13 @@ class CartMethod(base_synth.BaseSynthMethod):
         :param X: Feature dataset.
         :return: Synthesised target variable.
         """
+
+        if self.method_._all_missing:
+            return pd.Series(
+            np.nan,
+            index=X.index,
+            name=self.target_name_,
+        )
 
         check_is_fitted(self, ["method_", "feature_names_in_", "target_name_"])
 
--- a/tests/integration/test_int_bugfix_152.py
+++ b/tests/integration/test_int_bugfix_152.py
@@ -1,0 +1,34 @@
+"""
+This file is a placeholder for another issue:
+
+135-implement-more-integration tests
+
+This test is also present in the current version of that branch, where it failed.
+
+This file should be deleted when issue 135 merges with the develop branch
+"""
+
+from synthpop.synthesiser import Synthesiser
+from synthpop.methods.cart_synth import CartMethod
+from synthpop.methods.copy_synth import CopyMethod
+from synthpop.methods.sample_synth import SampleMethod
+
+import pandas as pd
+import numpy as np
+import pytest
+
+def test_int_bugfix_152():
+    df = pd.DataFrame({
+    "a": [1, None],
+    "b": [0, 0],
+    "c": [np.nan, np.nan]
+    })
+
+    special_syn_method = {
+    "a": SampleMethod(),
+    "b": CopyMethod(),
+    "c": CartMethod()
+    }
+
+    synth = Synthesiser(random_seed=2, special_syn_method=special_syn_method)
+    fit = synth.fit(df)
--- a/tests/integration/test_int_cart_method.py
+++ b/tests/integration/test_int_cart_method.py
@@ -262,3 +262,24 @@ def test_fit_transform_with_missing_values_in_target(y):
 
     assert len(out) == len(y)
     assert out.isna().sum() > 0
+
+    
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None], name='c')),
+        (pd.Series([np.nan, np.nan], name='c')),
+        (pd.Series([pd.NA, pd.NA], name='c'))
+    ]
+)
+def test_fitting_handles_entire_nan_array(y):
+    cart = CartMethod()
+
+    X = pd.DataFrame({
+        "a": [1, 2],
+        "b": [3, 4],
+    })
+
+    res = cart.fit(X, y)
+    assert res.method_._all_missing == True
+    assert res.target_name_ == 'c'
--- a/tests/unit/test_cart.py
+++ b/tests/unit/test_cart.py
@@ -12,6 +12,7 @@ from synthpop.utils import str_dtype
 class StubRegressor(TransformerMixin, BaseEstimator):
     def __init__(self, transform_result=None):
         self.transform_result = transform_result
+        self._all_missing = False
 
     def fit(self, X, y):
         self.fit_x = X
@@ -29,6 +30,7 @@ class StubRegressor(TransformerMixin, BaseEstimator):
 class StubClassifier(TransformerMixin, BaseEstimator):
     def __init__(self, transform_result=None):
         self.transform_result = transform_result
+        self._all_missing = False
 
     def fit(self, X, y):
         self.fit_x = X
--- a/tests/unit/test_cart.py
+++ b/tests/unit/test_cart.py
@@ -140,7 +140,9 @@ def test_fitting_handles_entire_nan_array(y):
 
     res = cart.fit(X, y)
     assert res.method_._all_missing == True
-    assert hasattr(res, 'target_name_')
+    assert res.target_name_ == 'c'
+    assert hasattr(res.method_, 'n_features_in_')
+    assert hasattr(res.method_, 'feature_order_')
     assert not hasattr(cart, "__sklearn_is_fitted__")
 
 
@@ -364,8 +366,6 @@ def test_transform_requires_fit():
     with pytest.raises(NotFittedError):
         cart.transform(pd.DataFrame({"a": [1]}))
 
-
-
 @pytest.mark.parametrize(
     "y",
     [
@@ -375,15 +375,19 @@ def test_transform_requires_fit():
 
     ]
 )
-def test_bypass_entire_nan_array(y):
-    cart = CartMethod()
+def test_transform_when_fitted_on_all_missing_outputs_all_missing(y):
+    cart = CartMethod(regressor=StubRegressor(), classifier=StubClassifier())
+
+    cart.method_ = StubRegressor()
+    cart.feature_names_in_ = ["a", "b"]
+    cart.target_name_ = "target"
 
     X = pd.DataFrame({
         "a": [1, 2, 3],
         "b": [4, 5, 6],
     })
 
-    cart.fit(X, y)
+    cart.method_._all_missing = True
     result = cart.transform(X)
 
     assert pd.isna(result).all()
--- a/tests/unit/test_cart.py
+++ b/tests/unit/test_cart.py
@@ -121,31 +121,6 @@ def test_fit_passes_standardised_data_to_tree(mocker):
     assert cart.method_.fit_x is clean_X
     assert cart.method_.fit_y is clean_y
 
-
-@pytest.mark.parametrize(
-    "y",
-    [
-        (pd.Series([None, None], name='c')),
-        (pd.Series([np.nan, np.nan], name='c')),
-        (pd.Series([pd.NA, pd.NA], name='c'))
-    ]
-)
-def test_fitting_handles_entire_nan_array(y):
-    cart = CartMethod()
-
-    X = pd.DataFrame({
-        "a": [1, 2],
-        "b": [3, 4],
-    })
-
-    res = cart.fit(X, y)
-    assert res.method_._all_missing == True
-    assert res.target_name_ == 'c'
-    assert hasattr(res.method_, 'n_features_in_')
-    assert hasattr(res.method_, 'feature_order_')
-    assert not hasattr(cart, "__sklearn_is_fitted__")
-
-
 @pytest.mark.parametrize(
     ("y", "expected_type"),
     [
--- a/tests/unit/test_cart.py
+++ b/tests/unit/test_cart.py
@@ -21,10 +21,10 @@ class StubRegressor(TransformerMixin, BaseEstimator):
     def transform(self, X):
         self.transform_x = X
         return self.transform_result
-    
+
     def get_feature_names_out(self, input_features=None):
         return ["fake_output"]
-    
+
 
 class StubClassifier(TransformerMixin, BaseEstimator):
     def __init__(self, transform_result=None):
@@ -38,7 +38,7 @@ class StubClassifier(TransformerMixin, BaseEstimator):
     def transform(self, X):
         self.transform_x = X
         return self.transform_result
-    
+
     def get_feature_names_out(self, input_features=None):
         return ["fake_output"]
 
@@ -48,8 +48,10 @@ class StubClassifier(TransformerMixin, BaseEstimator):
     ("X", "y", "expected", "message"),
     [
         ({}, pd.Series([1]), TypeError, "X must be a pandas DataFrame"),
-        (pd.DataFrame({"a": [1]}), [1], TypeError, "y must be a pandas Series"),
-        (pd.DataFrame({"a": [1, 2]}), pd.Series([1]), ValueError, "X and y must contain the same number of samples"),
+        (pd.DataFrame({"a": [1]}), [1],
+         TypeError, "y must be a pandas Series"),
+        (pd.DataFrame({"a": [1, 2]}), pd.Series(
+            [1]), ValueError, "X and y must contain the same number of samples"),
     ],
 )
 def test_fit_validates_inputs(X, y, expected, message):
@@ -57,6 +59,7 @@ def test_fit_validates_inputs(X, y, expected, message):
 
     with pytest.raises(expected, match=message):
         cart.fit(X, y)
+
 
 @pytest.mark.parametrize(
     ("y_array", "expected_type"),
@@ -88,6 +91,7 @@ def test_fit_selects_correct_method(mocker, y_array, expected_type):
     assert isinstance(cart.method_, expected_type)
     assert np.array_equal(cart.method_.fit_y, y_array)
 
+
 def test_fit_passes_standardised_data_to_tree(mocker):
     clean_X = {"a": np.array([1, 2, 3], dtype=np.float32)}
     clean_y = np.array([4, 5, 6], dtype=np.float32)
@@ -115,6 +119,29 @@ def test_fit_passes_standardised_data_to_tree(mocker):
     assert cart.method_.fit_x is clean_X
     assert cart.method_.fit_y is clean_y
 
+
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None], name='c')),
+        (pd.Series([np.nan, np.nan], name='c')),
+        (pd.Series([pd.NA, pd.NA], name='c'))
+    ]
+)
+def test_fitting_handles_entire_nan_array(y):
+    cart = CartMethod()
+
+    X = pd.DataFrame({
+        "a": [1, 2],
+        "b": [3, 4],
+    })
+
+    res = cart.fit(X, y)
+    assert res.method_._all_missing == True
+    assert hasattr(res, 'target_name_')
+    assert not hasattr(cart, "__sklearn_is_fitted__")
+
+
 @pytest.mark.parametrize(
     ("y", "expected_type"),
     [
@@ -132,11 +159,12 @@ def test_fit_clones_methods(y, expected_type):
     cart.fit(X, y)
 
     assert isinstance(cart.method_, expected_type)
-    
+
     if expected_type is StubRegressor:
         assert cart.method_ is not regressor
     else:
         assert cart.method_ is not classifier
+
 
 @pytest.mark.parametrize(
     "y",
@@ -179,15 +207,20 @@ def test_fit_sets_fitted_attributes(mocker, y):
     assert cart.method_.fit_y is clean_y
 
 # ----- transform tests -----
+
+
 @pytest.mark.parametrize(
     ("result", "method"),
     [
-        (pd.Series([10, 20]), StubRegressor(transform_result=np.array([10, 20]))),
-        (pd.Series(["a", "b"]), StubClassifier(transform_result=np.array(["a", "b"]))),
+        (pd.Series([10, 20]), StubRegressor(
+            transform_result=np.array([10, 20]))),
+        (pd.Series(["a", "b"]), StubClassifier(
+            transform_result=np.array(["a", "b"]))),
     ]
 )
 def test_transform_returns_series(mocker, result, method):
-    cart = CartMethod(regressor=StubRegressor(result), classifier=StubClassifier(result))
+    cart = CartMethod(regressor=StubRegressor(result),
+                      classifier=StubClassifier(result))
 
     cart.feature_names_in_ = ["a"]
     cart.target_name_ = "target"
@@ -216,15 +249,19 @@ def test_transform_returns_series(mocker, result, method):
 
     np.testing.assert_array_equal(out.to_numpy(), result)
 
+
 @pytest.mark.parametrize(
     ("result", "method"),
     [
-        (pd.Series([10, 20]), StubRegressor(transform_result=np.array([10, 20]))),
-        (pd.Series(["a", "b"]), StubClassifier(transform_result=np.array(["a", "b"]))),
+        (pd.Series([10, 20]), StubRegressor(
+            transform_result=np.array([10, 20]))),
+        (pd.Series(["a", "b"]), StubClassifier(
+            transform_result=np.array(["a", "b"]))),
     ]
 )
 def test_transform_preserves_metadata_and_feature_order(mocker, result, method):
-    cart = CartMethod(regressor=StubRegressor(result), classifier=StubClassifier(result))
+    cart = CartMethod(regressor=StubRegressor(result),
+                      classifier=StubClassifier(result))
 
     cart.method_ = method
     cart.feature_names_in_ = ["b", "a"]
@@ -262,6 +299,7 @@ def test_transform_preserves_metadata_and_feature_order(mocker, result, method):
         result,
     )
 
+
 def test_transform_rejects_missing_columns():
     cart = CartMethod(regressor=StubRegressor(), classifier=StubClassifier())
 
@@ -277,15 +315,19 @@ def test_transform_rejects_missing_columns():
     ):
         cart.transform(X)
 
+
 @pytest.mark.parametrize(
     ("result", "method"),
     [
-        (pd.Series([10, 20]), StubRegressor(transform_result=np.array([10, 20]))),
-        (pd.Series(["a", "b"]), StubClassifier(transform_result=np.array(["a", "b"]))),
+        (pd.Series([10, 20]), StubRegressor(
+            transform_result=np.array([10, 20]))),
+        (pd.Series(["a", "b"]), StubClassifier(
+            transform_result=np.array(["a", "b"]))),
     ]
 )
 def test_transform_ignores_extra_columns(mocker, result, method):
-    cart = CartMethod(regressor=StubRegressor(result), classifier=StubClassifier(result))
+    cart = CartMethod(regressor=StubRegressor(result),
+                      classifier=StubClassifier(result))
 
     cart.method_ = method
     cart.feature_names_in_ = ["b", "a"]
@@ -314,17 +356,45 @@ def test_transform_ignores_extra_columns(mocker, result, method):
         X[["b", "a"]],
     )
 
+
 def test_transform_requires_fit():
     cart = CartMethod()
     with pytest.raises(NotFittedError):
         cart.transform(pd.DataFrame({"a": [1]}))
 
+
+
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None, None])),
+        (pd.Series([np.nan, np.nan, np.nan])),
+        (pd.Series([pd.NA, pd.NA, pd.NA]))
+
+    ]
+)
+def test_bypass_entire_nan_array(y):
+    cart = CartMethod()
+
+    X = pd.DataFrame({
+        "a": [1, 2, 3],
+        "b": [4, 5, 6],
+    })
+
+    cart.fit(X, y)
+    result = cart.transform(X)
+
+    assert pd.isna(result).all()
+    assert len(result) == 3
 # ----- get_feature_names_out test -----
+
+
 def test_get_feature_names_out_delegates():
     cart = CartMethod()
     cart.method_ = StubRegressor()
 
     assert cart.get_feature_names_out() == ["fake_output"]
+
 
 def test_get_feature_names_out_raises_unfitted():
     cart = CartMethod()
@@ -333,6 +403,8 @@ def test_get_feature_names_out_raises_unfitted():
         cart.get_feature_names_out()
 
 # ----- clonability test -----
+
+
 def test_clone_works_and_fitted_cart_does_not_preserve_state():
     X = pd.DataFrame({"a": [1]})
     y = pd.Series([1])
--- a/tests/unit/test_tree_methods.py
+++ b/tests/unit/test_tree_methods.py
@@ -152,6 +152,7 @@ def fitted_tree(tree_method,request):
     tree_method.tree_sampler_ = tree_method.tree_sampler.clone()
     tree_method.tree_ = clone(tree_method.tree)
     tree_method.n_features_in_ = len(X.keys())
+    tree_method._all_missing = False
 
     tree_method.feature_order_ = list(X.keys())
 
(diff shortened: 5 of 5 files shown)


tests/unit/test_cart.py
```diff
@@ -115,6 +121,29 @@ def test_fit_passes_standardised_data_to_tree(mocker):
...
+
+    X = pd.DataFrame({
+        "a": [1, 2],
+        "b": [3, 4],
+    })
+
+    res = cart.fit(X, y)
+    assert res.method_._all_missing == True
+    assert hasattr(res, 'target_name_')
+    assert not hasattr(cart, "__sklearn_is_fitted__")
```
KoomenErina: Why should `cart` not have an `__sklearn_is_fitted__` attribute? What are you trying to assert?

tests/unit/test_cart.py
```diff
@@ -314,25 +358,55 @@ def test_transform_ignores_extra_columns(mocker, result, method):
...
+)
+def test_bypass_entire_nan_array(y):
+    cart = CartMethod()
+
+    X = pd.DataFrame({
+        "a": [1, 2, 3],
+        "b": [4, 5, 6],
+    })
+
+    cart.fit(X, y)
```
KoomenErina: I assume that this test is about `transform`. Calling `cart.fit` here makes this an integration test, not a unit test. 
Instead, manually set `cart` to the state that you expect it to be in when this method is called. In this scenario, it would involve `cart._all_missing=True`

tests/unit/test_cart.py
```diff
@@ -314,25 +358,55 @@ def test_transform_ignores_extra_columns(mocker, result, method):
...
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None, None])),
+        (pd.Series([np.nan, np.nan, np.nan])),
+        (pd.Series([pd.NA, pd.NA, pd.NA]))
+
+    ]
+)
+def test_bypass_entire_nan_array(y):
```
KoomenErina: if this test is about `transform`, it should be called something like `test_transform_bypass_entire_nan_array`
Even better would be `test_transform_when_fitted_on_all_missing_outputs_all_missing`

tests/unit/test_cart.py
```diff
@@ -115,6 +121,29 @@ def test_fit_passes_standardised_data_to_tree(mocker):
...
+    cart = CartMethod()
+
+    X = pd.DataFrame({
+        "a": [1, 2],
+        "b": [3, 4],
+    })
+
+    res = cart.fit(X, y)
+    assert res.method_._all_missing == True
+    assert hasattr(res, 'target_name_')
```
KoomenErina: `n_features_in_` and `feature_order_` should still be set.

KoomenErina: Now you only assert that `res` has a `target_name_`. It might be `None` and still pass this test. The name of the series in `y` is always 'c', so you can assert that `target_name_` is 'c' here

src/synthpop/methods/cart_synth.py
```diff
@@ -84,6 +84,11 @@ def fit(self, X: Dict[str, npt.NDArray], y: npt.NDArray) -> Self:
         self.target_name_ = getattr(y, "name", None)
         X_val, n_samples = utils.validate_2d_dict(X)
         y = utils.validate_1d_target(y, n_samples)
+        self._all_missing = False
```
KoomenErina, line 101: Why is the name `_all_missing` and not `all_missing_`? The first communicates that is an internal parameter. The second communicates that is an learned parameter.

src/synthpop/methods/cart_synth.py
```diff
@@ -414,6 +419,13 @@ def transform(self, X: pd.DataFrame) -> pd.Series:
         :return: Synthesised target variable.
         """
 
+        if hasattr(self, "method_") and getattr(self.method_, "_all_missing", False):
```
KoomenErina: I see that checking for all missing values is in `_AbstractTreeMethod` (fit), and that acting on that  (transform) is in `CartMethod`. The implementation of the rule "when every row of the target is missing, then all the rows of the output should be missing" is now spread of 2 classes. This complicates the code base: if you look at `_AbstractTreeMethod`, you don't see why `_all_missing` should be set. If you look at `CartMethod.transform`, you see that it depends on some internal/learned parameter. Depending on a internal parameter is not a good idea. Depending on the value of a learned parameter would be unusual in this context. 

Besides, this tightly couples `CartMethod` and `_AbstractTreeMethod`. a namechange in `_AbstractTreeMethod` would break `CartMethod`.

The reason for this bug is an artifact of that:
- sklearn decisiontrees don't accept missing values in the target
- MissingValuePredictor removes missing values

If we were to switch to XGBoost for example, we would not have this problem in the first place. If we were to remake `_AbstractTreeMethod` to use XGBoost, we would need to add a parameter which is not needed, but required by CartMethod to work. 

It is usually best to keep the implementation of this kind of rules close to the cause for the rules. That would mean that acting on `_all_missing` should happen in `_AbstractTreeMethod.transform`

KoomenErina (changes requested):
I don't see an integration test in test_int_cart_method.py that tests this behaviour. 

You mention problems with `check_is_fitted()`. What happend?

tests/unit/test_cart.py
casvanerpduo: As far as I understood _AbstractTreeMethod.fit() I now create a pathway that returns early, before any fitting of a decission tree has started. As such I assume cart has not fitted.

tests/unit/test_cart.py
casvanerpduo: self.method.feature_order_ is not set, as the fit() function in _AbstractTreeMethod is returned before that parameter is set. In the CartMethod.fit() self.feature_order_ does not exist

tests/unit/test_cart.py
```diff
@@ -115,6 +121,31 @@ def test_fit_passes_standardised_data_to_tree(mocker):
...
+
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None], name='c')),
+        (pd.Series([np.nan, np.nan], name='c')),
+        (pd.Series([pd.NA, pd.NA], name='c'))
+    ]
+)
+def test_fitting_handles_entire_nan_array(y):
```
KoomenErina: This is not a unit test. Calling `fit` here runs about the entire synthesis process. This can be made into an integration test by asserting just the correct output, and nothing more. (And ofcourse moving it to the correct file)

tests/unit/test_tree_methods.py
KoomenErina: There is no unit test that asserts setting the boolean in `fit` of the tree methods.
This unit test should also assert that none of the dependencies have been called. This can be done by asserting that the corresponding learned attributes (tree_, missing_handler_, etc...) do not exsist.

tests/unit/test_tree_methods.py
```diff
@@ -532,13 +534,41 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
...
+
+def test_transform_returns_nan_array_when_all_missing_true():
+    X = {
+        "a": np.array([1, 2, 3]),
+        "b": np.array([4, 5, 6]),
+    }
+
+    regressor = TreeRegressorMethod()
+
+    regressor._all_missing = True
```
KoomenErina: This rule is not just for `TreeRegressorMethod`, but also for `TreeClassifierMethod`. 
You can use `TestTreeMethod` like in the other unit tests. There is a fixture to assemble a fitted `TestTreeMethod`  (`fitted_tree`).

KoomenErina (changes requested):

tests/unit/test_tree_methods.py
```diff
@@ -518,53 +625,94 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
...
 
 
-#general tests ------------------------------------------------------------------------------------
+def test_transform_raises_not_fitted_when_missing_flag_absent():
+    X = {
+        "a": np.array([1, 2, 3]),
+        "b": np.array([4, 5, 6]),
+    }
+
+    regressor = TreeRegressorMethod()
```
KoomenErina: this tests only the `TreeRegressorMethod`

tests/unit/test_tree_methods.py
```diff
@@ -518,53 +625,94 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
...
+    with pytest.raises(NotFittedError):
+        regressor.transform(X)
+
+
+@pytest.mark.parametrize("X,index_cat", [(v[0], v[2]) for v in get_input_test_data()])
+def test_transform_returns_nan_array_when_all_missing_true(X, index_cat, fitted_tree):
+
+    fitted_tree._all_missing = True
+
+    result = fitted_tree.transform(X)
```
KoomenErina, line 721: test that none of the dependencies are used by setting them to None

tests/integration/test_int_cart_method.py
```diff
@@ -261,4 +261,25 @@ def test_fit_transform_with_missing_values_in_target(y):
...
+    
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None], name='c')),
+        (pd.Series([np.nan, np.nan], name='c')),
+        (pd.Series([pd.NA, pd.NA], name='c'))
+    ]
+)
+def test_fitting_handles_entire_nan_array(y):
```
KoomenErina: Add an integration test that includes `CartMethod.transform`

tests/unit/test_cart.py
```diff
@@ -314,25 +335,57 @@ def test_transform_ignores_extra_columns(mocker, result, method):
...
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None, None])),
+        (pd.Series([np.nan, np.nan, np.nan])),
+        (pd.Series([pd.NA, pd.NA, pd.NA]))
+
+    ]
+)
+def test_transform_when_fitted_on_all_missing_outputs_all_missing(y):
```
KoomenErina: This test succeeds in error.
This test should not be here, since the logic moved to AbstractTreeMethod. The test still passes because the stub returns None, and pandas puts that value on every index.

tests/integration/test_int_bugfix_152.py
```diff
@@ -0,0 +1,34 @@
...
+from synthpop.synthesiser import Synthesiser
+from synthpop.methods.cart_synth import CartMethod
+from synthpop.methods.copy_synth import CopyMethod
+from synthpop.methods.sample_synth import SampleMethod
+
+import pandas as pd
+import numpy as np
+import pytest
+
+def test_int_bugfix_152():
```
KoomenErina: This can be an integration test of synthesiser

KoomenErina (changes requested):

tests/unit/test_cart.py
```diff
@@ -314,25 +336,39 @@ def test_transform_ignores_extra_columns(mocker, result, method):
...
     )
 
+
 def test_transform_requires_fit():
     cart = CartMethod()
     with pytest.raises(NotFittedError):
         cart.transform(pd.DataFrame({"a": [1]}))
 
+
+@pytest.mark.parametrize(
```
KoomenErina: Did you run the tests in this file?

I could not run the tests in this file, since this decorator does not belong here.

tests/unit/test_tree_methods.py
```diff
@@ -36,478 +42,579 @@ def fit_transform(self, X, y=None, **fit_params):
...
 
-    assert np.array_equal(converted_y,tree_method.tree_.fit_y_)
+    assert np.array_equal(converted_y, tree_method.tree_.fit_y_)
     assert tree_method.tree_.fit_y_.dtype == np.float32
 
-    
 
-# test transform ----------------------------------------------------------------------------------
+@pytest.mark.parametrize("X,y,index_cat", get_input_test_data())
+def test_tree_method_fit_returns_nan_for_nan_array(X, y, index_cat, tree_method):
```
KoomenErina: the name is a bit misleading, since `fit` returns `self`.

tests/unit/test_tree_methods.py
```diff
@@ -518,53 +625,102 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
...
+    X = {
+        "a": np.array([1, 2, 3]),
+        "b": np.array([4, 5, 6]),
+    }
+
+    regressor = TreeRegressorMethod()
+    classifier = TreeClassifierMethod()
+
+    with pytest.raises(NotFittedError):
+        regressor.transform(X)
```
KoomenErina: this does not test `classifier.transform` if `regressor.transform` raises the expected error.

tests/unit/test_tree_methods.py
```diff
@@ -518,53 +625,102 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
...
+@pytest.mark.parametrize("X,index_cat", [(v[0], v[2]) for v in get_input_test_data()])
+def test_transform_returns_nan_array_when_all_missing_true(X, index_cat, fitted_tree):
+
+    expected = np.array([np.nan, np.nan, np.nan, np.nan, np.nan, np.nan])
+
+    fitted_tree._all_missing = True
+    
+    result = fitted_tree.transform(X)
+
+    tree_method.encoders_ = None
```
KoomenErina: `tree_method.<property> = None` does not do anything relevant here.
`tree_method` is a fixture, which is a function. If you use `tree_method` as the name of an argument for a test function, then pytest sets that argument to the returnvalue of that function. 

if you don't use `tree_method` as an argument, the symbol still exsists, but is a function.

Note that `fitted_tree.<property> = None` has no relevant effect here, since these lines are after the call to `transform`. 

If you set those propertief of `fitted_tree` to `None` before the call to `fitted_tree.transform`, then you effectivly test that those dependencies are not used in this test case.

tests/unit/test_tree_methods.py
```diff
@@ -656,17 +662,27 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
     assert result.dtype == str_dtype
 
 
-def test_transform_raises_not_fitted_when_missing_flag_absent():
+def test_regressor_transform_raises_not_fitted_when_missing_flag_absent():
```
KoomenErina: this test and ` test_classifier_transform_raises_not_fitted_when_missing_flag_absent` can be combined using the `fitted_tree` fixture, simular to ` test_transform_returns_nan_array_when_all_missing_true`

Also note that these tests ( `test_regressor_transform_raises_not_fitted_when_missing_flag_absent` and `test_classifier_transform_raises_not_fitted_when_missing_flag_absent` would pass even before dealing with a np.nan array was implementend. These tests make an unfitted tree method and call transform. That did already raise a NotFitted exception.

tests/unit/test_tree_methods.py
```diff
@@ -676,16 +692,20 @@ def test_transform_returns_nan_array_when_all_missing_true(X, index_cat, fitted_
...
 
-    tree_method.encoders_ = None
-    tree_method.missing_handler_ = None
-    tree_method.tree_sampler_ = None
-    tree_method.tree_ = None
-    tree_method.n_features_in_ = None
+    result = fitted_tree.transform(X)
 
     assert np.array_equal(result, expected, equal_nan=True)
+    assert fitted_tree.encoders_ == None
```
KoomenErina: `is None` is preferred to `== None`

tests/unit/test_tree_methods.py
```diff
@@ -676,16 +692,20 @@ def test_transform_returns_nan_array_when_all_missing_true(X, index_cat, fitted_
...
+    assert fitted_tree.missing_handler_ == None
```
KoomenErina: asserting that all dependencies are `None` after a transform is not needed.

tests/integration/test_int_synthesiser.py
```diff
@@ -161,3 +163,30 @@ def test_synthesiser_preserves_cat_cat_relation():
     for col in obs_ct.columns:
         value = np.max(np.abs(obs_ct[col] - syn_ct[col]))
         assert value < 0.1
+
+def test_int_bugfix_152():
```
anneliesdej: I would change the title to something like `test_synthesiser_handles_cart_with_all_missing_target` and then in thenI'd add a docstring to this test that explains that it is a regression test for bug 152 with a few words on what the bug was

src/synthpop/methods/cart_synth.py
```diff
@@ -119,6 +125,12 @@ def transform(self, X: Dict[str, npt.NDArray]) -> npt.NDArray:
 
         """
 
+        if not hasattr(self, '_all_missing'):
+            raise NotFittedError
+        elif self._all_missing:
+            return np.array([np.nan]*len(X[next(iter(X))]))
```
anneliesdej: I don't love this line. It exists before the input validation but assumes at least one feature for X. Safer (for silly inputs) would be:
```python
n = len(next(iter(X.values()))) if X else 0
return np.full(n, np.nan, dtype=np.float32)
```

I also changed the dtype in this suggestion because the current return is float64

src/synthpop/methods/cart_synth.py
```diff
@@ -119,6 +125,12 @@ def transform(self, X: Dict[str, npt.NDArray]) -> npt.NDArray:
 
         """
 
+        if not hasattr(self, '_all_missing'):
+            raise NotFittedError
+        elif self._all_missing:
```
anneliesdej, line 151: This new behaviour is undocumented now. To the docstrings I would add one sentence like:
If the target consists entirely of missing values, the estimator records this state and returns without fitting a tree. Subsequent calls to transform() produce an all-missing output

Although, I see that the MissingValuePredictor also lacks this so it might not be very important to add

src/synthpop/methods/cart_synth.py
```diff
@@ -119,6 +125,12 @@ def transform(self, X: Dict[str, npt.NDArray]) -> npt.NDArray:
 
         """
 
+        if not hasattr(self, '_all_missing'):
+            raise NotFittedError
+        elif self._all_missing:
+            return np.array([np.nan]*len(X[next(iter(X))]))
+
+
```
anneliesdej:
```suggestion 
```
One blank line is fine

src/synthpop/methods/cart_synth.py
anneliesdej:
```suggestion
all_features_dict = {
    k: self.encoders_[k].transform(v)
    if k in self.encoders_
    else v
    for (k, v) in X_val.items()
}
```
Just seeing now how hard it is to read

tests/unit/test_tree_methods.py
```diff
@@ -518,53 +631,95 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
...
-    assert np.array_equal(result,tree_method.missing_handler_.post_synth_transform_result)
+    assert np.array_equal(
+        result, tree_method.missing_handler_.post_synth_transform_result
+    )
     assert result.dtype == str_dtype
 
+@pytest.mark.parametrize("X,index_cat", [(v[0], v[2]) for v in get_input_test_data()])
+def test_transform_raises_not_fitted_when_missing_flag_absent(X, index_cat, fitted_tree):
+
+    delattr(fitted_tree, "_all_missing")
```
anneliesdej: It would be better to just start with a clean nonfitted tree instead of deleting the attribute

tests/unit/test_tree_methods.py
```diff
@@ -36,478 +42,585 @@ def fit_transform(self, X, y=None, **fit_params):
...
-    
 
-# test transform ----------------------------------------------------------------------------------
+@pytest.mark.parametrize("X,y,index_cat", get_input_test_data())
+def test_tree_method_fit_returns_no_fitting_parameters_for_nan_array(X, y, index_cat, tree_method):
+    y = np.array([np.nan, np.nan, np.nan, np.nan, np.nan, np.nan])
 
+    fitted = tree_method.fit(X, y)
+
+    assert fitted._all_missing
```
anneliesdej, line 524: There is now a test that checks if it is correctly set to True, but not one to see if it is correctly set to False

tests/unit/test_tree_methods.py
```diff
@@ -36,478 +42,585 @@ def fit_transform(self, X, y=None, **fit_params):
...
+    tree_method.fit(X, y)
+
+    assert np.array_equal(converted_y, tree_method.tree_.fit_y_)
     assert tree_method.tree_.fit_y_.dtype == np.float32
 
-    
 
-# test transform ----------------------------------------------------------------------------------
+@pytest.mark.parametrize("X,y,index_cat", get_input_test_data())
+def test_tree_method_fit_returns_no_fitting_parameters_for_nan_array(X, y, index_cat, tree_method):
```
anneliesdej: The returned dtype isn't tested for the all-missing path. The regressor should produce float32 and the classifier the str_dtype

anneliesdej:
```suggestion
def test_tree_method_fit_returns_no_fitted_parameters_for_nan_array(X, y, index_cat, tree_method):
```

tests/unit/test_tree_methods.py
```diff
@@ -518,53 +631,95 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
...
+    fitted_tree.tree_sampler_ = None
+    fitted_tree.tree_ = None
+    fitted_tree.n_features_in_ = None
 
-#general tests ------------------------------------------------------------------------------------
+    result = fitted_tree.transform(X)
 
-@pytest.mark.parametrize("X",[v[0] for v in get_input_test_data()])
-def test_get_feature_names_out(X,tree_method):
+    assert np.array_equal(result, expected, equal_nan=True)
```
anneliesdej:
```suggestion
    assert np.isnan(result).all()
    assert len(result) == 6
```
This is easier to use

tests/integration/test_int_cart_method.py
```diff
@@ -264,4 +264,47 @@ def test_fit_transform_with_missing_values_in_target(y):
     out = cart.transform(X)
 
     assert len(out) == len(y)
-    assert out.isna().sum() > 0
\ No newline at end of file
+    assert out.isna().sum() > 0
+
+    
+@pytest.mark.parametrize(
```
anneliesdej, line 271: Add an integration test on dtype return

tests/integration/test_int_cart_method.py
```diff
@@ -264,4 +264,47 @@ def test_fit_transform_with_missing_values_in_target(y):
...
+def test_fitting_handles_entire_nan_array(y):
+    cart = CartMethod()
+
+    X = pd.DataFrame({
+        "a": [1, 2],
+        "b": [3, 4],
+    })
+
+    res = cart.fit(X, y)
+    assert res.method_._all_missing == True
```
anneliesdej:
```suggestion
    assert res.method_._all_missing is True
```

tests/integration/test_int_cart_method.py
```diff
@@ -264,4 +264,47 @@ def test_fit_transform_with_missing_values_in_target(y):
...
+    X = pd.DataFrame({
+        "a": [1, 2],
+        "b": [3, 4],
+    })
+
+    res = cart.fit(X, y)
+    out = res.transform(X)
+    
+    assert len(out) == len(y)
+    assert pd.isna(out).all()
```
anneliesdej, line 311: I'd also add
```python
assert out.name == "c"
```
Because this is what users see

tests/integration/test_int_cart_method.py
```diff
@@ -264,4 +264,47 @@ def test_fit_transform_with_missing_values_in_target(y):
...
+
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None], name='c')),
+        (pd.Series([np.nan, np.nan], name='c')),
+        (pd.Series([pd.NA, pd.NA], name='c'))
+    ]
+)
+def test_transform_handles_entire_nan_array(y):
```
anneliesdej:
```suggestion
def test_transform_handles_all_missing_target(y):
```

tests/integration/test_int_cart_method.py
```diff
@@ -264,4 +264,47 @@ def test_fit_transform_with_missing_values_in_target(y):
...
+    
+@pytest.mark.parametrize(
+    "y",
+    [
+        (pd.Series([None, None], name='c')),
+        (pd.Series([np.nan, np.nan], name='c')),
+        (pd.Series([pd.NA, pd.NA], name='c'))
+    ]
+)
+def test_fitting_handles_entire_nan_array(y):
```
anneliesdej:
```suggestion
def test_fitting_handles_all_missing_target(y):
```

tests/integration/test_int_synthesiser.py
anneliesdej:
```suggestion
```
Improt does not get used

tests/integration/test_int_synthesiser.py
```diff
@@ -161,3 +163,30 @@ def test_synthesiser_preserves_cat_cat_relation():
...
+    assert isinstance(fit.models_['a'], SampleMethod)
+    assert isinstance(fit.models_['b'], CopyMethod)
+    assert isinstance(fit.models_['c'], CartMethod)
+
+    generated = fit.generate()
+
+    assert df['b'].equals(generated['b'])
+
+    for col in df.columns:
+        assert df[col].isin(generated[col]).all()
```
anneliesdej: This assertion is a little weak. For column `c` it only checks `np.nan in generated["c"]`. I would change it to `assert generated["c"].isna().all()` as that is the regression condition.
I would also add:
`assert len(generated) == len(df)` so that a future change in generation size does not hide the problem. The order would then be:
```python
generated = fit.generae()

assert len(generated) == len(df)

assert df['b'].equals(generated['b'])

assert generated['c'].isna().all()
```

tests/integration/test_int_synthesiser.py
```diff
@@ -161,3 +163,30 @@ def test_synthesiser_preserves_cat_cat_relation():
     for col in obs_ct.columns:
         value = np.max(np.abs(obs_ct[col] - syn_ct[col]))
         assert value < 0.1
+
+def test_int_bugfix_152():
+    df = pd.DataFrame({
+    "a": [1, None],
+    "b": [0, 0],
+    "c": [np.nan, np.nan]
```
anneliesdej: I would add to this regression test that replacing np.nan with None or pd.NA does not produce an error. You could parametrize it with 
```python
@pytest.mark.parametrize(
    "missing_value",
    [np.nan, None, pd.NA],
)
```
and then change:
```python
"c": [missing_value, missing_value]
```

tests/integration/test_int_synthesiser.py
```diff
@@ -161,3 +163,30 @@ def test_synthesiser_preserves_cat_cat_relation():
     for col in obs_ct.columns:
         value = np.max(np.abs(obs_ct[col] - syn_ct[col]))
         assert value < 0.1
+
+def test_int_bugfix_152():
+    df = pd.DataFrame({
```
anneliesdej: The lines below are not indented properly

tests/integration/test_int_synthesiser.py
```diff
@@ -161,3 +163,30 @@ def test_synthesiser_preserves_cat_cat_relation():
...
+    "a": [1, None],
+    "b": [0, 0],
+    "c": [np.nan, np.nan]
+    })
+
+    special_syn_method = {
```
anneliesdej, line 184: here are also the lines below not indented properly

tests/integration/test_int_synthesiser.py
```diff
@@ -161,3 +163,41 @@ def test_synthesiser_preserves_cat_cat_relation():
     for col in obs_ct.columns:
         value = np.max(np.abs(obs_ct[col] - syn_ct[col]))
         assert value < 0.1
+
+@pytest.mark.parametrize(
+    "missing_value",
+    [np.nan, None, pd.NA],
+)
+def test_synthesiser_handles_cart_with_all_missing_target(missing_value):
+    """This test resulted in bugfix 152. The CartMethod failed for a numerical entire np.nan array,
```
anneliesdej:
```suggestion
    """Regression test for bug 152. The CartMethod failed for a numerical array with only missing values, 
```

tests/integration/test_int_synthesiser.py
```diff
@@ -161,3 +163,41 @@ def test_synthesiser_preserves_cat_cat_relation():
...
         value = np.max(np.abs(obs_ct[col] - syn_ct[col]))
         assert value < 0.1
+
+@pytest.mark.parametrize(
+    "missing_value",
+    [np.nan, None, pd.NA],
+)
+def test_synthesiser_handles_cart_with_all_missing_target(missing_value):
+    """This test resulted in bugfix 152. The CartMethod failed for a numerical entire np.nan array, 
+    as columns with nan are masked. As such, fitting on an entire np.nan array is the same as fitting to an empty array.
```
anneliesdej:
```suggestion
    Missing values are masked. As such, fitting on an entire np.nan array is the same as fitting on an empty array.
```

tests/integration/test_int_synthesiser.py
```diff
@@ -161,3 +163,41 @@ def test_synthesiser_preserves_cat_cat_relation():
...
+    Which failed. A work-around wass implemented"""
```
anneliesdej:
```suggestion
    This threw an error resulting in bug issue 152. A work-around was implemented"""
```

src/synthpop/methods/cart_synth.py
```diff
@@ -111,14 +121,24 @@ def fit(self, X: Dict[str, npt.NDArray], y: npt.NDArray) -> Self:
 
     def transform(self, X: Dict[str, npt.NDArray]) -> npt.NDArray:
         """
-        Synthesise new column
+        Synthesise new column.
+
+        If the target consists entirely of missing values, 
+        the estimator records this state and returns without fitting a tree. 
+        Subsequent calls to transform() produce an all-missing output
```
anneliesdej: This text should be in fit, not transform

tests/integration/test_int_synthesiser.py
```diff
@@ -1,12 +1,13 @@
 import string
```
anneliesdej:
```suggestion
```
Unused import

src/synthpop/methods/cart_synth.py
```diff
@@ -127,9 +136,9 @@ def transform(self, X: Dict[str, npt.NDArray]) -> npt.NDArray:
         if not hasattr(self, '_all_missing'):
             raise NotFittedError
         elif self._all_missing:
-            return np.array([np.nan]*len(X[next(iter(X))]))
+            n = len(next(iter(X.values()))) if X else 0
+            return np.full(n, np.nan) # __super__() in specific method transforms will do correct dtype conversion
```
KoomenErina: 'method in child class' is a more appropiate wording.

tests/integration/test_int_synthesiser.py
```diff
@@ -205,17 +164,27 @@ def test_synthesiser_preserves_cat_cat_relation():
...
-    "c": [np.nan, np.nan]
-    })
+@pytest.mark.parametrize(
+    "missing_value",
+    [np.nan, None, pd.NA],
+)
+def test_synthesiser_handles_cart_with_all_missing_target(missing_value):
+    """This test resulted in bugfix 152. The CartMethod failed for a numerical entire np.nan array, 
+    as columns with nan are masked. As such, fitting on an entire np.nan array is the same as fitting to an empty array.
+    Which failed. A work-around wass implemented"""
```
KoomenErina: A "work-around" sounds like a temporarly fix while the root cause is not yet resolved. The root cause is that we did not specify what should happen when the target is fully nan.  We have specified it in this pull request. I would say this is a fix, and not a work-around.

src/synthpop/methods/cart_synth.py
```diff
@@ -11,6 +11,7 @@
 from sklearn.base import BaseEstimator, TransformerMixin, check_is_fitted
 from sklearn.decomposition import PCA
 from sklearn.tree import BaseDecisionTree, DecisionTreeClassifier, DecisionTreeRegressor
+from sklearn.exceptions import NotFittedError
```
anneliesdej, line 17: This one should be below the sklearn.decomposition alphabetically for PEP8. The synthpop imports are also not ordered alphabetically

src/synthpop/methods/cart_synth.py
anneliesdej, line 174:
```suggestion
```
One less white line

tests/integration/test_int_synthesiser.py
```diff
@@ -1,12 +1,11 @@
-import string
-
 import numpy as np
 import pandas as pd
 import pytest
-from sklearn.exceptions import NotFittedError
 
 from synthpop.synthesiser import Synthesiser
 from synthpop.methods.cart_synth import CartMethod
+from synthpop.methods.sample_synth import SampleMethod
```
anneliesdej: Order alphabetically

src/synthpop/methods/cart_synth.py
```diff
@@ -115,14 +137,20 @@ def fit(self, X: Dict[str, npt.NDArray], y: npt.NDArray) -> Self:
...
+        Synthesise new column.
 
         :param X: features used to predict the target variable.
 
         :return: synthesised column.
 
         """
 
+        if not hasattr(self, '_all_missing'):
+            raise NotFittedError
```
anneliesdej:
```suggestion
            raise NotFittedError(f"{self.__class__.__name__} instance is not fitted yet.")
```

src/synthpop/methods/cart_synth.py
```diff
@@ -2,25 +2,34 @@
 This module contains the CART method for synthesising data.
```
anneliesdej, line 2: Can you autopep8 this file?

tests/unit/test_tree_methods.py
```diff
@@ -544,53 +654,102 @@ def test_classifier_transform_returns_str_dtype(leafnode_sampler):
...
-    assert np.array_equal(result,tree_method.missing_handler_.post_synth_transform_result)
+    assert np.array_equal(
+        result, tree_method.missing_handler_.post_synth_transform_result
+    )
     assert result.dtype == str_dtype
 
 
-#general tests ------------------------------------------------------------------------------------
+@pytest.mark.parametrize("X,index_cat", [(v[0], v[2]) for v in get_input_test_data()])
+def test_transform_raises_not_fitted_when_missing_flag_absent(X, index_cat, tree_method):
```
anneliesdej:
```suggestion
def test_transform_raises_not_fitted_when_missing_flag_absent(X, index_cat, tree_method, missing_handling, leafnode_sampler):
```

tests/unit/test_tree_methods.py
```diff
@@ -367,61 +408,75 @@ def test_fit_tree_is_fit(X,y,index_cat,tree_method,mocker):
...
-    tree_method.fit(X,y)
+    tree_method.fit(X, y)
 
     assert tree_method.target_name_ is None
 
-def test_fit_classifier_converts_to_str(encoder,leafnode_sampler,mocker):
-    X = {"a":np.array([1,2])}
-    y = np.array(["a","b"],dtype=str_dtype)
 
+def test_fit_classifier_converts_to_str(encoder, leafnode_sampler, mocker):
```
anneliesdej, line 457: This test duplicates half of itself. Looks like remnants of a merge conflict

tests/unit/test_tree_methods.py
anneliesdej:
```suggestion
    expected_y = tree_method._convert_y(tree_method.missing_handler_.prepared_for_fit_result[1])
```
For the regressor, the previous one is correct, however for the classifier it is not necessarily true because `_convert_y()` converts to fixed-length unicode. This expected value tests the intended behaviour

anneliesdej:
```suggestion
    mock_fit_decision_tree_with_reachable_leaves.mock_calls.assert_called_once(), "_fit_decision_tree_with_reachable_leaves should be called 1 time"
```

anneliesdej:
```suggestion
    kwargs = mock_fit_decision_tree_with_reachable_leaves.call_args.kwargs
```

anneliesdej:
```suggestion
    actual_y = mock_fit_decision_tree_with_reachable_leaves.call_args.kwargs["y"]
```

tests/unit/test_tree_methods.py
```diff
@@ -367,61 +408,75 @@ def test_fit_tree_is_fit(X,y,index_cat,tree_method,mocker):
...
+        tree=StubTree(),
+    )
 
-    tree_method.fit(X,y)
+    tree_method.fit(X, y)
 
     mocked_to_str.assert_called_with(y)
-
+    
     actual_y = mock_fit_decision_tree_with_reachable_leaves.mock_calls[0][2]["y"]
```
anneliesdej:
```suggestion
    actual_y = mock_fit_decision_tree_with_reachable_leaves.call_args.kwargs["y"]
```

tests/unit/test_tree_methods.py
```diff
@@ -367,61 +408,75 @@ def test_fit_tree_is_fit(X,y,index_cat,tree_method,mocker):
...
     assert np.array_equal(str_y,actual_y)
```
anneliesdej:
```suggestion
    assert np.array_equal(str_y, actual_y)
    assert str_y.dtype == actual_y.dtype
```

tests/unit/test_tree_methods.py
```diff
@@ -9,11 +9,17 @@
 from sklearn.exceptions import NotFittedError
```
anneliesdej, line 9: Can you autopep8 this file?

tests/unit/test_cart.py
```diff
@@ -179,15 +187,20 @@ def test_fit_sets_fitted_attributes(mocker, y):
...
-        (pd.Series(["a", "b"]), StubClassifier(transform_result=np.array(["a", "b"]))),
+        (pd.Series([10, 20]), StubRegressor(
+            transform_result=np.array([10, 20]))),
+        (pd.Series(["a", "b"]), StubClassifier(
+            transform_result=np.array(["a", "b"]))),
     ]
 )
 def test_transform_returns_series(mocker, result, method):
-    cart = CartMethod(regressor=StubRegressor(result), classifier=StubClassifier(result))
+    cart = CartMethod(regressor=StubRegressor(result),
```
anneliesdej:
```suggestion
    cart = CartMethod()
```
method is already set in cart.method_ = method

tests/unit/test_cart.py
anneliesdej:
```suggestion
    mocked_X.assert_called_once_with(X)
```

anneliesdej, line 267: For some reason the suggestion won't take with multiple rows now, but this assert can be changed to:
mocked_standardise.assert_called_once_with(X[["b", "a"]])

casvanerpduo: That raises the error:    raise ValueError(
    ...<2 lines>...
    )
ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all()

